PR #10, opened 2024-04-27
[SPARK-48017]Add Spark application submission worker for operator

jiangzho:

This is a breakdown PR of #2  - adding a submission worker implementation for SparkApplication.

Spark Operator needs a submission worker to convert its abstraction (the SparkApplication API) into k8s resource spec.
This is a light-weight implementation based on native k8s integration. 

As of now, it's based off Spark3.5.1 - but it's assumed to serve all Spark LTS versions. This is feasible because as it aims to cover only the spec generation, Spark core jars are still brought-in by application images. E2Es would set up with operator later to ensure that. 

Per SPIP doc, in future operator version(s) we may add more implementations for submission worker based on different Spark versions to achieve 100% version agnostic, at the cost of having multiple workers stand-by. 

No

Added unit test coverage.

no


gradle.properties
```diff
@@ -18,17 +18,23 @@
 group=org.apache.spark.k8s.operator
 version=0.1.0
 
-fabric8Version=6.12.1
+# Caution: fabric8 version should be aligned with Spark dependency
+fabric8Version=6.7.2
```
jiangzho, line 22: This is caused by the co-exist dependency of Spark 3.5.1 .  fabric8 `6.7.2` and `6.12.1` are not binary compatible, it complains even if we exclude that from Spark dependency. 

A reminder has been added here to align the fabric8 version used by Spark and operator. I'll update this when 4.0.0 rcs become available

jiangzho: The incompatible was captured via unit tests.

dongjoon-hyun (changes requested):
Thank you for making a PR.

As we observe here, I'm not sure if the AS-IS design is sufficient or robust across multiple Spark versions. If we cannot find a way to be robust, it will bite us in the future again and again in the production.

It seems that we need a shim layer or multiple module like Iceberg
- https://github.com/apache/iceberg/tree/main/spark/v3.3
- https://github.com/apache/iceberg/tree/main/spark/v3.4
- https://github.com/apache/iceberg/tree/main/spark/v3.5

jiangzho:
> It seems that we need a shim layer or multiple module like Iceberg

Yes, that's one mid-term goal that we will target for operator v1.0, in order to achieve fully version agnostic.

This PR proposes single submission worker based on latest spark-kubernetes - consider it's history, we tested the compatibility with Spark 3.2, 3.3, 3.4, 3.5. We can do the same for 4.0 to ensure no breaking change is introduced. This is the pattern adopted by most operator solutions, like Flink operator / Google Spark operator. I'm not saying this is the absolutely right way to go for longer term, but it could enable the first batch of evaluations on operator 0.1 while we work on the multi-submission worker mode.

The challenges of multi-version submission worker mode involves 

* the operator image can be heavy (packaging multiple Spark jars)
* runtime resource consumption can be higher, because we need multiple containers (per Spark version) to avoid jar conflicts in class path.
* deployment (helm chart) of operator can be a bit more complex when users are more familiar with operator. i.e., users could choose to deploy operator with single submission worker mode, or a selection of Spark versions, or all known versions based on the need.

Given this can we start with this PR for v0.1 ?

dongjoon-hyun:
Thanks. Let me consider more.

dongjoon-hyun:
Thank you for renaming the package and updating LICENSE.

dongjoon-hyun (commented):
Apache Spark 4.0.0-preview will arrive next Monday. Let's resume this PR with that as the first Spark version until we have version-agnostic capability.

gradle.properties
```diff
@@ -18,17 +18,23 @@
...
 commonsLang3Version=3.14.0
 commonsIOVersion=2.16.1
 lombokVersion=1.18.32
 
+#Spark
+scalaVersion=2.12
```
dongjoon-hyun, line 28: As you know, there is not Scala 2.12 from Apache Spark 4.0.0-preview.

So, I guess we need to use Scala 2.13 only in Spark Operator for both case Scala 2.12 and 2.13 submission.

spark-submission-worker/build.gradle
```diff
@@ -0,0 +1,18 @@
+dependencies {
+  implementation project(":spark-operator-api")
+
+  implementation("org.apache.spark:spark-kubernetes_$scalaVersion:$sparkVersion")
```
dongjoon-hyun, line 4: Let me try to define the used classes as `DeveloperAPI` for this work. That will protect `spark-kubernetes-operator` from any accidental change from 4.0.0.

For now, Apache Spark guarantees only the following.
```
$ git grep '@DeveloperApi' resource-managers/kubernetes/
resource-managers/kubernetes/core/src/main/scala/org/apache/spark/deploy/k8s/KubernetesUtils.scala:@DeveloperApi
resource-managers/kubernetes/core/src/main/scala/org/apache/spark/deploy/k8s/SparkKubernetesClientFactory.scala:@DeveloperApi
resource-managers/kubernetes/core/src/main/scala/org/apache/spark/deploy/k8s/SparkPod.scala:@DeveloperApi
resource-managers/kubernetes/core/src/main/scala/org/apache/spark/deploy/k8s/features/KubernetesDriverCustomFeatureConfigStep.scala:@DeveloperApi
resource-managers/kubernetes/core/src/main/scala/org/apache/spark/deploy/k8s/features/KubernetesExecutorCustomFeatureConfigStep.scala:@DeveloperApi
resource-managers/kubernetes/core/src/main/scala/org/apache/spark/deploy/k8s/features/KubernetesFeatureConfigStep.scala:@DeveloperApi
resource-managers/kubernetes/core/src/main/scala/org/apache/spark/scheduler/cluster/k8s/AbstractPodsAllocator.scala:@DeveloperApi
resource-managers/kubernetes/core/src/main/scala/org/apache/spark/scheduler/cluster/k8s/ExecutorPodsPollingSnapshotSource.scala:@DeveloperApi
resource-managers/kubernetes/core/src/main/scala/org/apache/spark/scheduler/cluster/k8s/ExecutorPodsWatchSnapshotSource.scala:@DeveloperApi
```

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppDriverConf.java
```diff
@@ -0,0 +1,66 @@
...
+import scala.Option;
+
+import org.apache.spark.SparkConf;
+import org.apache.spark.deploy.k8s.Config;
+import org.apache.spark.deploy.k8s.KubernetesDriverConf;
+import org.apache.spark.deploy.k8s.KubernetesVolumeUtils;
+import org.apache.spark.deploy.k8s.submit.KubernetesClientUtils;
+import org.apache.spark.deploy.k8s.submit.MainAppResource;
+
+public class SparkAppDriverConf extends KubernetesDriverConf {
```
dongjoon-hyun, line 31: Since `KubernetesDriverConf` is a private class, always there has been a breaking change on `KubernetesDriverConf` like the following.
- #46276

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppDriverConf.java
```diff
@@ -0,0 +1,66 @@
...
+
+  public static SparkAppDriverConf create(
+      SparkConf sparkConf,
+      String appId,
+      MainAppResource mainAppResource,
+      String mainClass,
+      String[] appArgs,
+      Option<String> proxyUser) {
+    // pre-create check only
+    KubernetesVolumeUtils.parseVolumesWithPrefix(
```
dongjoon-hyun, line 50: Ditto. `KubernetesVolumeUtils` is a private object which is not protected.

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppDriverConf.java
```diff
@@ -0,0 +1,66 @@
...
+  /** Application managed by operator has a deterministic prefix */
+  @Override
+  public String resourceNamePrefix() {
+    return sparkConf().getOption(Config.KUBERNETES_DRIVER_POD_NAME_PREFIX().key()).isEmpty()
+        ? appId()
+        : sparkConf().get(Config.KUBERNETES_DRIVER_POD_NAME_PREFIX().key());
+  }
+
+  public String configMapNameDriver() {
+    return KubernetesClientUtils.configMapName(String.format("spark-drv-%s", resourceNamePrefix()));
```
dongjoon-hyun, line 64: Ditto. `KubernetesClientUtils` is a private object.

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppSubmissionWorker.java
```diff
@@ -0,0 +1,106 @@
...
+        effectiveSparkConf.set(confKey, applicationSpec.getSparkConf().get(confKey));
+      }
+    }
+    if (MapUtils.isNotEmpty(confOverrides)) {
+      for (Map.Entry<String, String> entry : confOverrides.entrySet()) {
+        effectiveSparkConf.set(entry.getKey(), entry.getValue());
+      }
+    }
+    effectiveSparkConf.set("spark.kubernetes.namespace", app.getMetadata().getNamespace());
+    MainAppResource primaryResource = new JavaMainAppResource(Option.empty());
```
dongjoon-hyun, line 66: This one will be okay because `MainAppResource` is a trait although it's `private`. Let me double-check this.

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppResourceSpec.java
```diff
@@ -0,0 +1,127 @@
...
+
+import io.fabric8.kubernetes.api.model.Container;
+import io.fabric8.kubernetes.api.model.ContainerBuilder;
+import io.fabric8.kubernetes.api.model.HasMetadata;
+import io.fabric8.kubernetes.api.model.Pod;
+import io.fabric8.kubernetes.api.model.PodBuilder;
+import lombok.Getter;
+import org.apache.commons.lang3.StringUtils;
+
+import org.apache.spark.deploy.k8s.Config;
```
dongjoon-hyun, line 38: `Config` will be okay.

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppResourceSpec.java
```diff
@@ -0,0 +1,127 @@
...
+import org.apache.spark.deploy.k8s.Constants;
```
dongjoon-hyun, line 39: We need to add a note or warning to keep this `Constants` to be append-only mode from Apache Spark repository.

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppResourceSpec.java
```diff
@@ -0,0 +1,127 @@
...
+import org.apache.spark.deploy.k8s.KubernetesDriverSpec;
```
dongjoon-hyun, line 40: ditto. `KubernetesDriverSpec` is a private case class.

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppResourceSpec.java
```diff
@@ -0,0 +1,127 @@
...
+import org.apache.spark.deploy.k8s.SparkPod;
```
dongjoon-hyun, line 41: We are safe because `SparkPod` is defined as `DeveloperApi`.

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppResourceSpec.java
```diff
@@ -0,0 +1,127 @@
...
+      return;
+    }
+    resource.getMetadata().setNamespace(namespace);
+  }
+
+  private SparkPod addConfigMap(SparkPod pod, Map<String, String> confFilesMap) {
+    Container containerWithVolume =
+        new ContainerBuilder(pod.container())
+            .addNewEnv()
+            .withName(org.apache.spark.deploy.k8s.Constants.ENV_SPARK_CONF_DIR())
```
dongjoon-hyun, line 103: Do you need a full package name when we have `import` statement? Maybe, the following?
```java
- .withName(org.apache.spark.deploy.k8s.Constants.ENV_SPARK_CONF_DIR())
+ .withName(Constants.ENV_SPARK_CONF_DIR())
```

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppResourceSpec.java
```diff
@@ -0,0 +1,127 @@
...
+            .withValue(org.apache.spark.deploy.k8s.Constants.SPARK_CONF_DIR_INTERNAL())
```
dongjoon-hyun, line 104: ditto.

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppResourceSpec.java
```diff
@@ -0,0 +1,127 @@
...
+            .endEnv()
+            .addNewVolumeMount()
+            .withName(org.apache.spark.deploy.k8s.Constants.SPARK_CONF_VOLUME_DRIVER())
```
dongjoon-hyun, line 107: ditto.

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppResourceSpec.java
```diff
@@ -0,0 +1,127 @@
...
+            .withMountPath(org.apache.spark.deploy.k8s.Constants.SPARK_CONF_DIR_INTERNAL())
```
dongjoon-hyun, line 108: ditto.

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppDriverConf.java
dongjoon-hyun: Here is a PR for this.
- https://github.com/apache/spark/pull/46326

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppDriverConf.java
dongjoon-hyun: Here is the PR.
- https://github.com/apache/spark/pull/46327

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppResourceSpec.java
dongjoon-hyun: Here is a PR.
- https://github.com/apache/spark/pull/46329

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppSubmissionWorker.java
dongjoon-hyun: Here is a PR.
- https://github.com/apache/spark/pull/46332

dongjoon-hyun (commented):
FYI, Apache Spark 4.0.0-preview1 RC1 is here.

- https://repository.apache.org/content/repositories/orgapachespark-1454/org/apache/spark/spark-kubernetes_2.13/4.0.0-preview1/

gradle.properties
jiangzho: yup! upgrading to 2.13 along with 4.0.0-preview1

dongjoon-hyun:
Thank you for updating.

gradle.properties
```diff
@@ -18,17 +18,23 @@
...
 
+# Caution: fabric8 version should be aligned with Spark dependency
 fabric8Version=6.12.1
 commonsLang3Version=3.14.0
 commonsIOVersion=2.16.1
 lombokVersion=1.18.32
 
+#Spark
+scalaVersion=2.13
+sparkVersion=4.0.0-preview1
```
dongjoon-hyun, line 29: Thank you. Please update the following PR description and double-check the other description too.
> As of now, it's based off Spark3.5.1 - but it's assumed to serve all Spark LTS versions.

jiangzho: Updated, thanks for the catch!

The pull build currently fails fetching preview build - I can add a init gradle script to add additional maven repository as a temporary workaround. I'm also open if there's alternative recommended way from Spark community, please let me know.

jiangzho: @dongjoon-hyun - can we use rc1 for submission worker for now, in order to unblock the operator module ?

I'll be more than happy to upgrade & fix possible incompatibility when RC2 is ready.

dongjoon-hyun: > @dongjoon-hyun - can we use rc1 for submission worker for now, in order to unblock the operator module ?
> 
> I'll be more than happy to upgrade & fix possible incompatibility when RC2 is ready.

Thank you, @jiangzho . Thank you for updating to use RC1 first.

build.gradle
```diff
@@ -25,6 +25,11 @@ subprojects {
 
   repositories {
     mavenCentral()
+    // This is a workaround to resolve Spark 4.0.0-preview-1
```
dongjoon-hyun, line 28: `4.0.0-preview-1` -> `4.0.0-preview1`

- https://github.com/apache/spark/releases/tag/v4.0.0-preview1-rc1

build.gradle
```diff
@@ -25,6 +25,11 @@ subprojects {
...
+    // To be removed for official release
```
dongjoon-hyun, line 29: Apache Spark recommends to use `IDed TODO`s. Please file a JIRA issue and use it like the following.

```
- // To be removed for official release
+ // TODO(SPARK-XXXXX) Use Apache Spark 4.0.0-preview1 when it's ready
```

spark-operator-api/src/main/java/org/apache/spark/k8s/operator/utils/ModelUtils.java
```diff
@@ -107,4 +108,12 @@ public static boolean overrideExecutorTemplateEnabled(ApplicationSpec applicatio
         && applicationSpec.getExecutorSpec() != null
         && applicationSpec.getExecutorSpec().getPodTemplateSpec() != null;
   }
+
+  public static long getAttemptId(final SparkApplication app) {
```
dongjoon-hyun, line 112: Thank you for adding `final`.

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppSubmissionWorker.java
```diff
@@ -0,0 +1,147 @@
...
+import org.apache.spark.k8s.operator.utils.ModelUtils;
+
+/**
+ * Similar to org.apache.spark.deploy.k8s.submit.KubernetesClientApplication. This reads args from
+ * SparkApplication instead of starting separate spark-submit process
+ */
+public class SparkAppSubmissionWorker {
+  // Default length limit for generated app id. Generated id is used as resource-prefix when
+  // user-provided id is too long for this purpose. This applied to all resources associated with
+  // the Spark app (including k8s service which has different naming length limit). This we
```
dongjoon-hyun, line 49: `This we`?

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppSubmissionWorker.java
```diff
@@ -0,0 +1,147 @@
...
+      primaryResource = new JavaMainAppResource(Option.apply(applicationSpec.getJars()));
+      effectiveSparkConf.setIfMissing("spark.jars", applicationSpec.getJars());
+    } else if (StringUtils.isNotEmpty(applicationSpec.getPyFiles())) {
+      primaryResource = new PythonMainAppResource(applicationSpec.getPyFiles());
+      effectiveSparkConf.setIfMissing("spark.submit.pyFiles", applicationSpec.getPyFiles());
+    } else if (StringUtils.isNotEmpty(applicationSpec.getSparkRFiles())) {
+      primaryResource = new RMainAppResource(applicationSpec.getSparkRFiles());
+    }
+    effectiveSparkConf.setIfMissing(
+        "spark.master", "k8s://https://$KUBERNETES_SERVICE_HOST:$KUBERNETES_SERVICE_PORT");
```
dongjoon-hyun, line 90: To @jiangzho and @aaruna, as we know, Apache Spark's `ExternalClusterManager` allows a custom K8s-based external cluster manager. So, it would be great if `Spark K8s Operator` has a capability from the beginning to allow a custom prefix in addition to `k8s` from the beginning.

https://github.com/apache/spark/blob/51623785c38c9b17a6d91cb8e7f686459bd4803e/core/src/main/scala/org/apache/spark/scheduler/ExternalClusterManager.scala#L33
```scala
def canCreate(masterURL: String): Boolean
```

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppSubmissionWorker.java
```diff
@@ -0,0 +1,147 @@
...
+  public static String generateSparkAppId(final SparkApplication app) {
+    long attemptId = ModelUtils.getAttemptId(app);
+    String preferredId = String.format("%s-%d", app.getMetadata().getName(), attemptId);
+    if (preferredId.length() > DEFAULT_ID_LENGTH_LIMIT) {
+      int preferredIdPrefixLength =
+          DEFAULT_ID_LENGTH_LIMIT - DEFAULT_HASH_BASED_IDENTIFIER_LENGTH_LIMIT - 1;
+      String preferredIdPrefix = preferredId.substring(0, preferredIdPrefixLength);
+      return generateHashBasedId(
+          preferredIdPrefix,
+          app.getMetadata().getNamespace(),
```
dongjoon-hyun, line 123: This line seems to have some assumption on K8s namespace length.
- Can we support the maximum K8s namespace length here? 
- Please make a test case for `generateSparkAppId` for that maximum K8s namespace length.

spark-submission-worker/src/test/java/org/apache/spark/k8s/operator/SparkAppDriverConfTest.java
```diff
@@ -0,0 +1,50 @@
...
+    SparkConf sparkConf = new SparkConf();
+    sparkConf.set("foo", "bar");
+    sparkConf.set("spark.executor.instances", "1");
+    String appId = UUID.randomUUID().toString();
+    SparkAppDriverConf sparkAppDriverConf =
+        SparkAppDriverConf.create(
+            sparkConf, appId, mock(JavaMainAppResource.class), "foo", null, Option.empty());
+    String resourcePrefix = sparkAppDriverConf.resourceNamePrefix();
+    Assertions.assertEquals(resourcePrefix, appId);
+    Assertions.assertTrue(sparkAppDriverConf.configMapNameDriver().contains(resourcePrefix));
```
dongjoon-hyun, line 47: Please add more test cases seperately at least for all public methods. For example, it would be great if `configMapNameDriver` has a new test case.

spark-submission-worker/src/test/java/org/apache/spark/k8s/operator/SparkAppResourceSpecTest.java
```diff
@@ -0,0 +1,137 @@
...
+    Seq<HasMetadata> preResourceSeq = CollectionConverters.asScala(preResourceList).toList();
+    Seq<HasMetadata> resourceSeq = CollectionConverters.asScala(resourceList).toList();
+    when(mockSpec.driverKubernetesResources()).thenReturn(resourceSeq);
+    when(mockSpec.driverPreKubernetesResources()).thenReturn(preResourceSeq);
+    when(mockSpec.pod()).thenReturn(sparkPod);
+    when(mockSpec.systemProperties()).thenReturn(new HashMap<>());
+
+    SparkAppResourceSpec appResourceSpec = new SparkAppResourceSpec(mockConf, mockSpec);
+
+    Assertions.assertEquals(2, appResourceSpec.getDriverResources().size());
```
dongjoon-hyun, line 74: Shall we import `Assertions.assertEquals`?

spark-submission-worker/src/test/java/org/apache/spark/k8s/operator/SparkAppSubmissionWorkerTest.java
```diff
@@ -0,0 +1,197 @@
...
+      SparkApplication mockApp = mock(SparkApplication.class);
+      ApplicationSpec mockSpec = mock(ApplicationSpec.class);
+      ObjectMeta appMeta = new ObjectMetaBuilder().withName("app1").withNamespace("ns1").build();
+      Map<String, String> appProps = new HashMap<>();
+      appProps.put("foo", "bar");
+      appProps.put("spark.executor.instances", "1");
+      appProps.put("spark.kubernetes.namespace", "ns2");
+      Map<String, String> overrides = new HashMap<>();
+      overrides.put("spark.executor.instances", "5");
+      overrides.put("spark.kubernetes.namespace", "ns3");
```
dongjoon-hyun, line 66: Please add a new test coverage with a long namespace length.

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppResourceSpec.java
```diff
@@ -0,0 +1,129 @@
...
+            .addNewEnv()
+            .withName(Constants.ENV_SPARK_CONF_DIR())
+            .withValue(Constants.SPARK_CONF_DIR_INTERNAL())
+            .endEnv()
+            .addNewVolumeMount()
+            .withName(Constants.SPARK_CONF_VOLUME_DRIVER())
+            .withMountPath(Constants.SPARK_CONF_DIR_INTERNAL())
+            .endVolumeMount()
+            .build();
+    Pod podWithVolume =
```
dongjoon-hyun, line 112: It would be great to have more specific name like `podWithConfigMapVolume`.

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppSubmissionWorker.java
```diff
@@ -0,0 +1,147 @@
...
+    effectiveSparkConf.setIfMissing(
+        "spark.master", "k8s://https://$KUBERNETES_SERVICE_HOST:$KUBERNETES_SERVICE_PORT");
+    String appId = generateSparkAppId(app);
+    effectiveSparkConf.setIfMissing("spark.app.id", appId);
+    return SparkAppDriverConf.create(
+        effectiveSparkConf,
+        appId,
+        primaryResource,
+        applicationSpec.getMainClass(),
+        applicationSpec.getDriverArgs().toArray(new String[0]),
```
dongjoon-hyun, line 98: Please define a final constant in order to reuse this, `new String[0]`.

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppSubmissionWorker.java
```diff
@@ -0,0 +1,147 @@
...
+      return preferredId;
+    }
+  }
+
+  public static String generateHashBasedId(final String prefix, final String... identifiers) {
+    return generateHashBasedId(
+        prefix, DEFAULT_ENCODE_BASE, DEFAULT_HASH_BASED_IDENTIFIER_LENGTH_LIMIT, identifiers);
+  }
+
+  public static String generateHashBasedId(
```
dongjoon-hyun, line 136: This method looks a little unsafe because of varargs, `identifiers`.

- Could you provide a way to cap the maximum string length always?
- Also, please add a method description because this is a `public static` method.

dongjoon-hyun, line 136: Please add a method description because this is `public static` method. Especially, about the guaranteed return string length.

spark-submission-worker/src/test/java/org/apache/spark/k8s/operator/SparkAppSubmissionWorkerTest.java
```diff
@@ -0,0 +1,197 @@
...
+import org.apache.spark.SparkConf;
+import org.apache.spark.deploy.k8s.submit.JavaMainAppResource;
+import org.apache.spark.deploy.k8s.submit.PythonMainAppResource;
+import org.apache.spark.deploy.k8s.submit.RMainAppResource;
+import org.apache.spark.k8s.operator.spec.ApplicationSpec;
+import org.apache.spark.k8s.operator.status.ApplicationAttemptSummary;
+import org.apache.spark.k8s.operator.status.ApplicationStatus;
+import org.apache.spark.k8s.operator.status.AttemptInfo;
+
+class SparkAppSubmissionWorkerTest {
```
dongjoon-hyun, line 49: Please add more test coverage. For example, a corner case like `generateHashBasedId` whose input has long long identifiers.

dongjoon-hyun (commented):
Hi, @jiangzho . I finished another round of reviews. Could you address them?

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppSubmissionWorker.java
jiangzho: The namespace name / app name are used as values to generate hash which has a final length limit applied. Added Javadoc and one more test case to clarify this behavior.

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppSubmissionWorker.java
jiangzho: yes - it should be possible to use custom Cluster Manager by setting `spark.master`. When master is not expplicitly set in SparkConf, this would automatically generate master URL based on the environment which operator is deployed in.

To make it more handy (e.g. still support automatic generating master url but with a different prefix), a new property has been introduced. Added javaDoc in `getResourceSpec` to clarify the behavior

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppSubmissionWorker.java
jiangzho: thanks for the catch! fixed the typo.

dongjoon-hyun (approved):
+1, LGTM. Thank you for addressing all comments (with IDed TODOs).
Merged to main.

dongjoon-hyun:
I wrote the current status summary here.
- https://github.com/apache/spark-kubernetes-operator/pull/2#issuecomment-2120918277

PR #112, opened 2024-08-29
[SPARK-49453] Add quotes for number

TQJADE:
### What changes were proposed in this pull request?
[SPARK-49453] Add quotes for number in values.yaml

### Why are the changes needed?
I have another value.yaml as below:
``` yaml
operatorConfiguration:
  dynamicConfig:
  enable: true
  create: true
  data:
    spark.kubernetes.operator.watchedNamespaces: "default, spark-1"
```
After running below with two values.yaml:
helm install spark-kubernetes-operator --create-namespace -f build-tools/helm/spark-kubernetes-operator/values.yaml -f tests/e2e/helm/dynamic-config-values.yaml build-tools/helm/spark-kubernetes-operator/
The generated configmap data field does not contains the line spark.kubernetes.operator.watchedNamespaces: "default, spark-1". Note that if you run helm install --dry-run, the record exist

### Does this PR introduce _any_ user-facing change?
no

### How was this patch tested?
test locally

### Was this patch authored or co-authored using generative AI tooling?
no

PR #114, opened 2024-08-29
[SPARK-49465] Fix example props override in helm chart operatorConfiguration.dynamicConfig.data

jiangzho:

This PR updates example hot property override in values.yaml / `operatorConfiguration.dynamicConfig.data`, make it including key value pairs of strings

Number value in example may cause helm failure

```
Error: INSTALLATION FAILED: ConfigMap in version "v1" cannot be handled as a ConfigMap: json: cannot unmarshal number into Go struct field ConfigMap.data of type string
```

Also, in fact Spark Operator consumes all property override as string values, it would [parse the actual types accordingly](https://github.com/apache/spark-kubernetes-operator/blob/main/spark-operator/src/main/java/org/apache/spark/k8s/operator/config/ConfigOption.java#L81-L103).

value parsing covered by existing CI, helm installation tested via minikube setup and custom value override

No

### DIFF
--- a/build-tools/helm/spark-kubernetes-operator/values.yaml
+++ b/build-tools/helm/spark-kubernetes-operator/values.yaml
@@ -188,4 +188,4 @@ operatorConfiguration:
       "helm.sh/resource-policy": keep
     data:
       # Spark Operator Config Runtime Properties Overrides. e.g.
-      spark.kubernetes.operator.reconciler.intervalSeconds: 60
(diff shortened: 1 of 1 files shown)


dongjoon-hyun (commented):
This seems to be a duplication of
- #112

cc @TQJADE too

dongjoon-hyun:
This is superceded by #115

PR #135, opened 2024-09-23
[SPARK-49754] Support HPA for `SparkCluster`

dongjoon-hyun:

This PR aims to support K8s [Horizontal Pod Autoscaler](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale) for `SparkCluster`.

To allow users more flexible installation on top of static SparkClusters.

No. This is a new feature and HPA is not created if the minimum number of workers is equal to the maximum number of workers.

Pass the CIs. And, manually create an example cluster and wait until it scales down to the minimum number of workers.
```
$ gradle build buildDockerImage spark-operator-api:relocateGeneratedCRD
$ kubectl apply -f examples/cluster-with-hpa.yaml
```

```
Conditions:                                                                                                                                                          │
│   Type            Status  Reason            Message                                                                                                                  │
│   ----            ------  ------            -------                                                                                                                  │
│   AbleToScale     True    ReadyForNewScale  recommended size matches current size                                                                                    │
│   ScalingActive   True    ValidMetricFound  the HPA was able to successfully calculate a replica count from cpu resource utilization (percentage of request)         │
│   ScalingLimited  True    TooFewReplicas    the desired replica count is less than the minimum replica count                                                         │
│ Events:                                                                                                                                                              │
│   Type    Reason             Age    From                       Message                                                                                               │
│   ----    ------             ----   ----                       -------                                                                                               │
│   Normal  SuccessfulRescale  2m31s  horizontal-pod-autoscaler  New size: 2; reason: All metrics below target                                                         │
│   Normal  SuccessfulRescale  91s    horizontal-pod-autoscaler  New size: 1; reason: All metrics below target
```

No.

### DIFF
--- a/examples/cluster-with-hpa.yaml
+++ b/examples/cluster-with-hpa.yaml
@@ -1,0 +1,45 @@
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+apiVersion: spark.apache.org/v1alpha1
+kind: SparkCluster
+metadata:
+  name: cluster-with-hpa
+spec:
+  runtimeVersions:
+    sparkVersion: "4.0.0-preview2"
+  clusterTolerations:
+    instanceConfig:
+      initWorkers: 3
+      minWorkers: 1
+      maxWorkers: 3
+  workerSpec:
+    workerStatefulSetSpec:
+      template:
+        spec:
+          containers:
+          - name: worker
+            resources:
+              requests:
+                cpu: "3"
+                memory: "3Gi"
+              limits:
+                cpu: "3"
+                memory: "3Gi"
+  sparkConf:
+    spark.kubernetes.container.image: "apache/spark:4.0.0-preview2-java21"
+    spark.master.ui.title: "Cluster with HorizontalPodAutoscaler"
+    spark.master.rest.enabled: "true"
+    spark.master.rest.host: "0.0.0.0"
+    spark.ui.reverseProxy: "true"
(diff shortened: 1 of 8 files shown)


dongjoon-hyun:
Could you review this PR, @viirya ?

examples/cluster-with-hpa.yaml
```diff
@@ -0,0 +1,45 @@
...
+        spec:
+          containers:
+          - name: worker
+            resources:
+              requests:
+                cpu: "3"
+                memory: "3Gi"
+              limits:
+                cpu: "3"
+                memory: "3Gi"
```
viirya, line 39: Hmm, this definition is enough for hpa? Don't we need to specify scaling behavior?

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkClusterResourceSpec.java
```diff
@@ -271,4 +277,55 @@ private static StatefulSet buildWorkerStatefulSet(
...
+                new MetricSpecBuilder()
+                    .withType("Resource")
+                    .withNewResource()
+                    .withName("cpu")
+                    .withNewTarget()
+                    .withType("Utilization")
+                    .withAverageUtilization(30)
+                    .endTarget()
+                    .endResource()
+                    .build())
```
viirya, line 311: Can these configs like resource and average utilization to be specify in cluster yaml file instead hard-coded?

viirya (commented):
Should we also update the documentation to mention this support?

examples/cluster-with-hpa.yaml
dongjoon-hyun: To @viirya , this is not HPA part. This is a minimum pod spec which HPA requires later.

Previously, the user creates HPA manually, in this PR, HPA part is hard-coded here for now. It will be revised later by allowing template.
- https://github.com/apache/spark-kubernetes-operator/pull/135/files#diff-2366f4d987646a8a1cd7d555ab69d582f647bfd86c3fdc773506f6e2743b3da2R313-R326

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkClusterResourceSpec.java
dongjoon-hyun: Ya, template support is the next step, @viirya .

dongjoon-hyun:
> Should we also update the documentation to mention this support?

Not yet~ CPU-based HPA is just for illustrating the simplest integration. Eventually, we will finalize with a custom metric based on Spark utilization API and document all of them together.
- https://github.com/apache/spark/pull/44908

viirya (approved):

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkClusterResourceSpec.java
viirya: Okay. Looking forward to the next steps. Thanks.

dongjoon-hyun:
Thank you, @viirya !

examples/cluster-with-hpa.yaml
viirya: Okay. Then I expect that we will add more into this example after revising later.

dongjoon-hyun:
Merged to main.

PR #140, opened 2024-09-30
[SPARK-49831] Provide empty `RuntimeVersions` object to `ClusterSpec.runtimeVersions` by default

dongjoon-hyun:

Like the other fields of `ClusterSpec`, this PR aims to provide at least empty `RuntimeVersions` object to `ClusterSpec.runtimeVersions` field by default.

Although `ClusterSpec.runtimeVersions` field is a required field, it's `null` currently by default because this should be provided by the users. However, technically, we can create `RuntimeVersions` object with null fields when we don't know the values of `RuntimeVersion` class fields. By providing this intermediate object by default, we can write a test case more easily in other modules.

No.

Pass the CIs with the updated test case.

No.


dongjoon-hyun:
cc @viirya

viirya (approved):

dongjoon-hyun:

PR #159, opened 2025-02-28
[SPARK-51347] Enable Ingress and Service Support for Spark Driver

jiangzho:

This PR adds support for launching Ingress and Services with Spark Applications.

There's common ask from user to support exposing driver via ingress and service to serve HTTP / HTTPS from outside the cluster - for example, to access Spark UI.

As we support podTemplateSupport already, this feature can also be handy for those who would like to expose additional service ports in the same pod. Therefore, this PR aims to introduce this feature in a more general way instead of just exposing Spark UI.

No - unreleased version

CIs

No

### DIFF
--- a/docs/spark_custom_resources.md
+++ b/docs/spark_custom_resources.md
@@ -99,6 +99,46 @@ Please be advised that Spark still overrides necessary pod configuration in both
 more details,
 refer [Spark doc](https://spark.apache.org/docs/latest/running-on-kubernetes.html#pod-template).
 
+## Enable Additional Ingress for Driver
+
+Operator may create [Ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) for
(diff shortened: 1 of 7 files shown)


dongjoon-hyun:
Thank you for making a PR, @jiangzho .

dongjoon-hyun:
BTW, I noticed that this is a first commit with a different email, @jiangzho . Are you going to use this one?
```
$ git log --author=zh | grep 'Author:' | sort | uniq -c
   1 Author: Zhou JIANG <jiangzho@umich.edu>
  20 Author: zhou-jiang <zhou_jiang@apple.com>
```

jiangzho:
Thanks for the reminder! This is a result of local squashing, I'll fix this for following commits - will still use `jiangzho@umich.edu`

PR #166, opened 2025-03-14
[SPARK-51507] Support creating config map that can be mounted by Spark pods for apps

jiangzho:

This PR introduces a new field `configMapSpecs`, which enables user to create configmap(s) which can be later mounted on driver and/or executors with values specified in SparkApplications. It also adds corresponding docs and unit tests.

There's a common user demand for the feature to create lightweight values / files override using configmap. This new feature saves the trouble for user to create configmap(s) beforehand - the config map would share the lifecycle of the corresponding Spark app.

User may mount the ConfigMap as needed in driver and executor pod templates - one example is provided in the doc. In future we may further enhance this flow by mounting the created config map automatically to Spark pods. This can be achieved by supporting volume type of ConfigMap in Spark Kubernetes, or by adding mutating webhook & alter pod templates after creation. For now, this PR opens up the possibility of creating config map and mounting them with one single spec.

No - not yet released

Pass the CIs

No

### DIFF
--- a/docs/spark_custom_resources.md
+++ b/docs/spark_custom_resources.md
@@ -138,6 +138,62 @@ default rule backed by the associated Service. It's recommended to always provid
 to make sure it's compatible with your 
 [IngressController](https://kubernetes.io/docs/concepts/services-networking/ingress-controllers/).
 
+## Create and Mount ConfigMap
+
+It is possible to ask operator to create configmap so they can be used by driver and/or executor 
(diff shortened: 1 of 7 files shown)


dongjoon-hyun (approved):
+1, LGTM (Pending CIs).

dongjoon-hyun:
Merged to main. Thank you, @jiangzho .

PR #204, opened 2025-05-16
[SPARK-52173] Fix OwnerReference Patching for Spark Application Driver Resources

jiangzho:

This PR fixes the logic to patch ownerReference on pre-resources for application. It would decorate the proposed spec instead of created resource with generated fields.

At initialization phase, operator would request pre-resources for driver first, then apply the owner references on these resources with created driver pod uid.

The latter patching is currently failing as it's decorating on created resource (which have managed fields set).

```
Message: metadata.managedFields must be nil.
```

> ... when using the Apply operation you cannot define managedFields in the body of the request that you submit.

https://kubernetes.io/docs/reference/using-api/server-side-apply/

The patch body shall not include generated `managedFields` .

No

Unit tests added

No


dongjoon-hyun (approved):
+1, LGTM. Thank you, @jiangzho .
Merged to main.

PR #220, opened 2025-05-22
[SPARK-52253] Support both `v1alpha1` and `v1beta1`

dongjoon-hyun:

This PR aims to support `v1alpha1` and `v1beta1`.
- Both `v1alpha1` and `v1beta1` will be accepted and be stored as `v1beta1`.
- Since the CRD generator always write a single version, we cannot use it directly from now. We should keep all previous versions in the CRD files inevitably.
- It's also required to protect these CRDs from accidental and breaking changes from `Beta`.

The added CRD files look like the following. Please note that `v1alpha` is `storage: false` and `v1beta1` is `storage: true`.

```yaml
# ... APACHE HEADER ...
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: sparkapplications.spark.apache.org
spec:
  group: spark.apache.org
  names:
    kind: SparkApplication
    plural: sparkapplications
    shortNames:
      - sparkapp
    singular: sparkapplication
  scope: Namespaced
  versions:
    - name: v1alpha1
      storage: false
       ...
    - name: v1beta1
      storage: true
       ...
```

- To support a smooth migration.
- To be clear, there is no schema change from 0.2.0 to 0.3.0 so far. Only, versions are changed.

No behavior change. 

Pass the CIs.

After manual installation,

```
$ kubectl get crds sparkapplications.spark.apache.org -oyaml | yq '.spec.versions[].name'
v1alpha1
v1beta1

$ kubectl get crds sparkapplications.spark.apache.org -oyaml | yq .status.storedVersions
- v1beta1
```

No.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -118,7 +118,6 @@ jobs:
         run: |
           eval $(minikube docker-env)
           ./gradlew buildDockerImage
-          ./gradlew spark-operator-api:relocateGeneratedCRD
           helm install spark-kubernetes-operator --create-namespace -f build-tools/helm/spark-kubernetes-operator/values.yaml build-tools/helm/spark-kubernetes-operator/
           helm test spark-kubernetes-operator
           # Use remote host' s docker image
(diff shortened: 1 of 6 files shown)


dongjoon-hyun:
cc @viirya , @peter-toth , @jiangzho

viirya (approved):

dongjoon-hyun:
Thank you, @viirya ! Merged to main.

PR #246, opened 2025-06-14
[SPARK-52478] Use GCS `Maven Central` mirror as the primary repository

dongjoon-hyun:

This PR aims to use GCS `Maven Central` mirror as the primary repository.

Since Apache Spark 3.0.0, Apache Spark community has been taking advantage of GCS `Maven Central` to stabilize and speed up CIs. We need to follow this practice in Apache Spark K8s Operator repository.
- apache/spark#26793

In addition, this PR cleaned up the following which is a leftover of Apache Spark 4.0.0-SNAPSHOT usage before the official 4.0.0 release.

https://github.com/apache/spark-kubernetes-operator/blob/9705573c9beebc0607104ae288c3e443d7494fd4/build.gradle#L58-L60

No behavior change. This is a build-only dev-side change.

Pass the CIs.

No.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -61,9 +61,6 @@ subprojects {
       url = uri("https://maven-central.storage-download.googleapis.com/maven2/")
     }
     mavenCentral()
-    maven {
-      url = "https://repository.apache.org/content/repositories/snapshots/"
-    }
   }
 
   apply plugin: 'checkstyle'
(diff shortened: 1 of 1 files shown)


dongjoon-hyun:
Let me merge this since this is the same like Apache Spark.

PR #257, opened 2025-06-26
[SPARK-52581]Revise AppCleanUpStep to include cleanup logic for all states

jiangzho:

This PR updates AppCleanUpStep so that it can be added as a regular step of each reconciliation, remove the therefore-unused AppTerminatedStep

Previously, we have AppTerminatedStep performed at the beginning of each reconciliation, to end the actions early if app has reached terminated state.

However, in some scenarios, we may still want to do "clean up" for applications that are terminated:

* If an app in "terminated without releasing resources" is being deleted, we would like the AppCleanUpStep to be performed, since it covers a few corner cases that might not be covered in general delete flow
* We may also to introduce the concept of "maximal retention duration" for applications along with current "retention policy" in near future to serve the lifecycle management of application secondary resources. That could also require cleaning up secondary resources for already-terminated applications.

The revised AppCleanUpStep includes everything in the previous AppTerminatedStep and takes care of cleaning up resources for those in "TerminatedWithoutReleaseResources" stage as needed.

No - internal step(s) changes only

CIs and E2Es shall cover the clean up logic

No


dongjoon-hyun (approved):
Sorry for the delay. I used sick days for Thursday and Friday.

+1, LGTM. Thank you, @jiangzho . I'll merge this and test more.

dongjoon-hyun:
BTW, I'm preparing `v0.4.0` to deliver this, @jiangzho  . The vote will start on July 1st.

PR #279, opened 2025-07-08
[SPARK-52702] Support using latest generated CRD yaml in Helm charts

jiangzho:

This PR allows the generated yaml as latest version to be used by Helm charts, after merging with older versions

To avoid manually updating staged CRD yaml files when updating spark-operator-api

No

CIs

No


spark-operator-api/src/main/resources/finalize-generated-crd.sh
```diff
@@ -23,5 +23,15 @@
...
   yq -i '.spec.versions[0] += ({"additionalPrinterColumns": [{"jsonPath": ".status.currentState.currentStateSummary", "name": "Current State", "type": "string"}, {"jsonPath": ".metadata.creationTimestamp", "name": "Age", "type": "date"}]})' $f
+  filename=$(basename "$f")
+  template_file_path="${SCRIPT_PATH}/../../../../build-tools/crd-templates/${filename}"
+  target_dir="${SCRIPT_PATH}/../../../../build-tools/helm/spark-kubernetes-operator/crds"
+  mkdir -p "${target_dir}"
+  target_file_path="${target_dir}/${filename}"
+  yq eval-all '
+    select(fileIndex == 1) as $source
+    | select(fileIndex == 0)
+    | .spec.versions += [$source.spec.versions[0]]
```
dongjoon-hyun: Oh, is this general enough to handle all cases when we change the APIs?

jiangzho: The intention was to:

* keep old CRD version(s) that do not need new features in a template file, and
* append new version at the end of the template, then move the full generated YAML to the proper location of the chart.

As we still want supporting v1alpha1 & v1beta1, I can remove them from the template. In the next commit we can do

* keep a template file that includes an empty list of versions
  - deprecated version(s) can be move to it in future
* keep a list of supported versions, generate latest YAML for them, then apply them to the template. 

How does this sound ?

dongjoon-hyun: Is it a little over-engineering without much clear benefits, isn't it. If we leave it in the existing way,
1. It's logically simple because only a PR (which changes the specs) will add minimal and required lines.
2. For all CRDs (including the latest CRD), we can have a full traceability with Git too.

dongjoon-hyun:
cc @peter-toth

spark-operator-api/src/main/resources/finalize-generated-crd.sh
jiangzho: I understand there's still concern about this auto-generation. The automation was intended to reduce potential human error in maintaining the large yaml file. 

While this PR itself can be a bit big for that purpose, can we at least reduce this to a test scenario validating the automated generated yaml in build is fully contained in the checked-in CRD yaml?

jiangzho:
please refer the lightweight test at #283

PR #291, opened 2025-07-29
[SPARK-52997] Fixes wrong worker assignment if multiple clusters are deployed to the same namespace

schmaxXximilian:
Updated the podSelector for master and worker services to include both clusterRole and name labels.

Using only clusterRole caused service misrouting when multiple Spark clusters were deployed in the same namespace. Adding name ensures correct pod targeting.

No, this is an internal fix to service selectors.

Tested with multiple clusters in the same namespace. Verified each service only matched its own pods via kubectl describe service.
Also adapted unit tests to reflect new behaviour

Yes, PR metadata was assisted by AI, but code changes were made manually.

### DIFF
--- a/spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkClusterResourceSpec.java
+++ b/spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkClusterResourceSpec.java
@@ -21,7 +21,7 @@ package org.apache.spark.k8s.operator;
 
 import static org.apache.spark.k8s.operator.Constants.*;
 
-import java.util.Map;
+import java.util.Collections;
 import java.util.Optional;
 
 import scala.Tuple2;
@@ -117,12 +117,9 @@ public class SparkClusterResourceSpec {
         .endMetadata()
         .withNewSpecLike(serviceSpec)
         .withClusterIP("None")
-        .withSelector(
-            Map.of(
-                LABEL_SPARK_CLUSTER_NAME,
-                name,
-                LABEL_SPARK_ROLE_NAME,
-                LABEL_SPARK_ROLE_MASTER_VALUE))
+        .addToSelector(Collections.singletonMap(LABEL_SPARK_CLUSTER_NAME, name))
+        .addToSelector(
+            Collections.singletonMap(LABEL_SPARK_ROLE_NAME, LABEL_SPARK_ROLE_MASTER_VALUE))
         .addNewPort()
         .withName("web")
         .withPort(8080)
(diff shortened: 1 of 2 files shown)


schmaxXximilian:
Hey @dongjoon-hyun, would be great if you could have a look at this. Our team is using the operator and we are grateful to the great work you folks have been doing.
Alas, the issue described in the PR is a pretty big problem to it. It would be amazing if we could get it resolved for the next release of the operator.

dongjoon-hyun:
Thank you for making a PR, @schmaxXximilian .

cc @peter-toth

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkClusterResourceSpec.java
```diff
@@ -117,8 +117,13 @@ private static Service buildMasterService(
         .endMetadata()
         .withNewSpecLike(serviceSpec)
         .withClusterIP("None")
+        .withClusterIP("None")
```
dongjoon-hyun: Why do we duplicate this, @schmaxXximilian ?

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkClusterResourceSpec.java
```diff
@@ -117,8 +117,13 @@ private static Service buildMasterService(
...
         .withSelector(
-            Collections.singletonMap(LABEL_SPARK_ROLE_NAME, LABEL_SPARK_ROLE_MASTER_VALUE))
+            Map.of(
```
dongjoon-hyun: May I ask why we need to use `Map.of` instead of `Collections.singletonMap`?

dongjoon-hyun (changes requested):
I left a few comments including https://github.com/apache/spark-kubernetes-operator/pull/291/files#r2243410142 .

In general, I understand your requirements although this is not recommended for `HPA`-enabled Spark Clusters. Let me play with this for a while.

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkClusterResourceSpec.java
schmaxXximilian: Hi @dongjoon-hyun, I'm afraid, I'm not quire sure what you are suggesting. `Collections.singletonMap` only creates a map with a single element. 
I changed my initial commit to below approach. Do you think it would be more suitable?
```
        .addToSelector(Collections.singletonMap(LABEL_SPARK_CLUSTER_NAME, name))
        .addToSelector(
            Collections.singletonMap(LABEL_SPARK_ROLE_NAME, LABEL_SPARK_ROLE_MASTER_VALUE))
```

dongjoon-hyun:
I added you to the Apache Spark contributor group (of ASF JIRA) and assigned SPARK-52997 to you.

Welcome to the Apache Spark community.

PR #296, opened 2025-08-08
[SPARK-53186] Fix probe port override from helm chart

jiangzho:

This is a bug fix to apply the health probe port from helm chart value override onto operator config

Without this patch, the port override - if  set in value.yaml - is only applied to operator container spec but not applied in config, would cause probe failures thereafter

No

CIs, helm lint, and local testing

No


jiangzho:
cc @peter-toth

peter-toth:
@jiangzho , can you please resove the conflict?

peter-toth:
Thanks @jiangzho for the fix!

Merged to `main` (0.5.0).

PR #299, opened 2025-08-27
[SPARK-53405] Add metrics recording for latency of Spark app state transition

jiangzho:

This PR adds metrics tracking state transition latency for Spark Application in format of

```
sparkapp.latency.from.<fromState>.to.<toState>
```

Latency measuring would be useful to analyze the performance from scheduling / orchestration perspective. For example, to analyze which state causes significant overhad and therefore optimize at cluster/app level.

More metrics becomes available.

CIs. New unit test added.

No

### DIFF
--- a/spark-operator/src/main/java/org/apache/spark/k8s/operator/metrics/source/OperatorJosdkMetrics.java
+++ b/spark-operator/src/main/java/org/apache/spark/k8s/operator/metrics/source/OperatorJosdkMetrics.java
@@ -23,12 +23,9 @@ import static io.javaoperatorsdk.operator.api.reconciler.Constants.CONTROLLER_NA
 
 import java.util.Map;
 import java.util.Optional;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.TimeUnit;
 
-import com.codahale.metrics.Counter;
 import com.codahale.metrics.Gauge;
-import com.codahale.metrics.Histogram;
 import com.codahale.metrics.MetricRegistry;
 import io.fabric8.kubernetes.api.model.HasMetadata;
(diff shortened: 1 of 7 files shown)


jiangzho:
cc @peter-toth for review

peter-toth:
Thanks @jiangzho for the PR. I am travelling for the rest of the week, but I can review this PR once I'm back.

spark-operator/src/main/java/org/apache/spark/k8s/operator/metrics/source/OperatorJosdkMetrics.java
```diff
@@ -104,14 +98,17 @@ public void receivedEvent(Event event, Map<String, Object> metadata) {
           getResourceClass(metadata);
       final Optional<String> namespaceOptional = event.getRelatedCustomResourceID().getNamespace();
       resource.ifPresent(
-          aClass -> getCounter(aClass, action.name().toLowerCase(), RESOURCE, EVENT).inc());
+          aClass ->
+              getCounter(getMetricName(aClass, action.name().toLowerCase(), RESOURCE, EVENT))
```
peter-toth: As you extracted `BaseOperatorSource` and refactored `getCounter()` / `getGauge()` / `getTimer()` to have only one `metricName` argument, you need to call `getMetricName()` almost as many times as you call the above APIs.
I wonder if you could keep an old methods to avoid the changes at these callsites.

If you really need the new, one argument versions, then you can add them as overloaded methods.

peter-toth:
The really like the idea, but let's try polish the implementation a bit.

peter-toth:
Thanks @jiangzho for the PR.

Merged to `main` (0.5.0).

PR #305, opened 2025-09-02
[SPARK-53457] Expose fabric8 dependencies as api for downstream users

jiangzho:

This PR exposes fabric8 dependencies as api in gradle build and hence in pom

Our public API surfaces fabric8 types (e.g. PodTemplateSpec), so hiding the dependencies behind `implementation` may cause comile errors for customers who need to use those types along with our API library.

This enhances developer experience and avoid surprises when integrating our API with projects that already depends on fabric8.

Pom change only

CIs

No

### DIFF
--- a/docs/spark_operator_api_library.md
+++ b/docs/spark_operator_api_library.md
@@ -47,5 +47,5 @@ your classpath when you depend on our library. It's recommended if you want to c
 Spark Operator API using Fabric8 types (for example, embedding a Fabric8 PodTemplateSpec into
 Spark Operator CRD).
 
-Do not depend on both `spark-operator-api` and `spark-operator-api-fabric8` at the same time. 
+Do not depend on both `spark-operator-api` and `spark-operator-api-fabric8` at the same time.
 Choose one based on your use case.


jiangzho:
cc @peter-toth for review - thanks!

dongjoon-hyun (changes requested):
Could you elaborate on the real example of the cases? AFAIK, the official users of this repositories are supposed to use or provide YAML files only, @jiangzho . Maybe, did you want to say some downstream fork? Exposing something should be very careful.

> Our public API surfaces fabric8 types (e.g. PodTemplateSpec), so hiding the dependencies behind implementation may cause comile errors for customers who need to use those types along with our API library.

jiangzho:
In addition to direct YAML, user may also leverage our Java API library to create and manage Spark workloads. Its a common practice for user to use a microservice to orchestrate and record status at higher level. ([example](https://github.com/apple/batch-processing-gateway) - they invented a set of POJO for Kubeflow operator previously as it's in Golang - but for Apache Spark Operator people may directly ask for Java API). While they build Spark spec in such microservices, they could be interacting with fabric8 models. It would become necessary for us to declare our dependencies in pom.

dongjoon-hyun:
We don't provide such a promise as a `Contract` yet, because it means that the exposed `fabric8` dependency will be a breaking change at every upgrade. Could you provide safe alternatives, @jiangzho ?

> In addition to direct YAML, user may also leverage our Java API library to create and manage Spark workloads. Its a common practice for user to use a microservice to orchestrate and record status at higher level. ([example](https://github.com/apple/batch-processing-gateway) - they invented a set of POJO for Kubeflow operator previously as it's in Golang - but for Apache Spark Operator people may directly ask for Java API). While they build Spark spec in such microservices, they could be interacting with fabric8 models. It would become necessary for us to declare our dependencies in pom.

jiangzho:
>  Could you provide safe alternatives ?

To avoid breaking changes in our current artifact - how about publishing one additional artifact, like `spark-operator-api-fabric8` to serve the Java users / microservices ? We can declare dependencies in the pom for the new artifact

dongjoon-hyun:
For the record, I landed a new fabric8 dependency which is marked as a breaking change from their side.
- #316 
- https://github.com/fabric8io/kubernetes-client/releases/tag/v7.4.0

dongjoon-hyun:
For the record, I'm currently auditing the dependencies officially. This kind of activity should be done before exposing anything officially in the ASF community. Thank you for raising this kind of use-case issue, @jiangzho . 
- #332
- #346 
- #345 
- #344 
- #343 
- #342

dongjoon-hyun:
BTW, please make the PR title and description up-to-date according to your changes (proposing a new java library module), @jiangzho .
> spark-operator-api-fabric8

docs/spark_operator_api_library.md
```diff
@@ -0,0 +1,51 @@
...
+"AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+KIND, either express or implied.  See the License for the
+specific language governing permissions and limitations
+under the License.
+-->
+
+# Use Spark Operator API Library in Your Java Application
+
+In addition to direct YAML, you may also leverage our Java API library to create and manage Spark
+workloads.
```
dongjoon-hyun, line 23: Please mention that Apache Spark community doesn't aim to provide a compatibility for this module in terms of `Semantic Versioning`. Every version can talk to the exactly same version's `Spark Operator`, @jiangzho .

PR #322, opened 2025-09-17
[SPARK-53628] Upgrade `PMD/SpotBugs` and enable checking on Java 25

dongjoon-hyun:

This PR aims to upgrade `pmd` and `spotbugs` test dependency and enable it on Java 25.

Previously, `PMD` and `SpotBugs` check are enabled only on Java 17 and 21.

No behavior change because this is a code quality check during building.

Pass the CIs. Java 25 test pipeline will cover this.

No.

### DIFF
--- a/config/pmd/ruleset.xml
+++ b/config/pmd/ruleset.xml
@@ -41,6 +41,7 @@
     <exclude name="TooManyStaticImports" />
     <exclude name="UseExplicitTypes" />
     <exclude name="UseUnderscoresInNumericLiterals" />
+    <exclude name="TypeParameterNamingConventions" />
   </rule>
   <rule ref="category/java/design.xml">
     <exclude name="AbstractClassWithoutAnyMethod" />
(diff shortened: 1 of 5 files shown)


config/pmd/ruleset.xml
```diff
@@ -41,6 +41,7 @@
     <exclude name="TooManyStaticImports" />
     <exclude name="UseExplicitTypes" />
     <exclude name="UseUnderscoresInNumericLiterals" />
+    <exclude name="TypeParameterNamingConventions" />
```
dongjoon-hyun, line 44: Although this is a new rule, it doesn't fit our usage. For example, we use *word* for parameter names but this rule enforces to use a single letter. It doesn't improve the code readability in our case.

https://github.com/apache/spark-kubernetes-operator/blob/c07e61eaf17b785777b49a078efe085f18164d3a/spark-operator-api/src/main/java/org/apache/spark/k8s/operator/BaseResource.java#L34-L45

dongjoon-hyun:

PR #336, opened 2025-09-24
[SPARK-53701] Use Java 25 toolchain while supporting Java 17+

dongjoon-hyun:
### What changes were proposed in this pull request?

This PR aims to start to use `Java 25 (Azul Zulu JDK)` toolchain while supporting Java 17+ from `Apache K8s Operator v0.5`.

### Why are the changes needed?

After this PR, we can take advantage of Java 25 compiler improvements while supporting all Java 17+ versions via `options.release = 17`.

https://github.com/apache/spark-kubernetes-operator/blob/db7e6403fd14ab86c302532423e26d3799a6e4c8/build.gradle#L59

### Does this PR introduce _any_ user-facing change?

- There is no behavior change for users.
- There is no developer experience change in general.
   - For the system without Java 25, `Gradle` will automatically download and use it.
   - Since we assume that `Gradle Wrapper` downloads `Gradle` itself, most of developers should get Java 25 without any issues.

### How was this patch tested?

Pass the CIs.

### Was this patch authored or co-authored using generative AI tooling?

No.

PR #359, opened 2025-10-01
[SPARK-53778] Use JDK25 Gradle Image as builder images

dongjoon-hyun:

This PR aims to use JDK25 Gradle Image instead of JDK17.

Since SPARK-53701, Apache Spark K8s Operator has been using JDK25 Toolchain which means it downloads JDK 25 inside JDK 17 Docker image. By using JDK25 Gradle Image, we can eliminate this redundancy and save resources during Docker images.

- #336 

No behavior change.

Pass the CIs.

No.

### DIFF
--- a/build-tools/docker/Dockerfile
+++ b/build-tools/docker/Dockerfile
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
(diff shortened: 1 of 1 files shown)


dongjoon-hyun:
Since this is a simple builder image change, I'm going to merge this.

PR #402, opened 2025-10-22
[SPARK-53992] Add `SparkOperatorConfManager.getAll` method

dongjoon-hyun:

This PR aims to add `SparkOperatorConfManager.getAll` method to get all configurations.

Like Apache Spark `SparkConf.getAll` API, we had better support `getAll` API.

https://github.com/apache/spark/blob/0be5f96e7a7d1b9fb9f96f1914019f79ef973a25/core/src/main/scala/org/apache/spark/SparkConf.scala#L197

```scala
/** Get all parameters as a list of pairs */
def getAll: Array[(String, String)]
```

No because this is a new API.

No.


dongjoon-hyun:
Let me merge this because this is a new API.

dongjoon-hyun:
Merged to main.

PR #422, opened 2025-12-10
[SPARK-54659] Enhance gradle task 'generateConfPropsDoc' to account for recent changes

jiangzho:

Enhance existing task `generateConfPropsDoc` and add new task `validateConfPropsDocUpToDate`

* Task `generateConfPropsDoc` would no longer automatically overwrite existing doc file. It would generate file for validation purpose only, and would perform overwrite only if property flag is enabled.
* Also fixed the file output location to use project rootDir instead of property value of `user.dir` as the latter may not point to project root in various cases.
* Added new task `validateConfPropsDocUpToDate` to compare the staged docs vs desired (generated) docs, and prompts developer to update docs if needed.

Although we have `generateConfPropsDoc` chained in the build task, it's not mandatory and hence docs can become out of date. By adding this patch, developers can be more mindful for doc updates corresponding to code change.

No. Building tasks update only.

Self-contained upon success & manually tested failure scenario when docs become out of date:

```
FAILURE: Build failed with an exception.

* Where:
Build file '/Users/zhou_jiang/work/bench/apache/ipr/apache-spark-kubernetes-operator/build-tools/docs-utils/build.gradle' line: 59

* What went wrong:
Execution failed for task ':build-tools-docs-utils:validateConfPropsDocUpToDate'.
> Doc /Users/zhou_jiang/work/bench/apache/ipr/apache-spark-kubernetes-operator/docs/config_properties.md out of date: if operator conf properties has been updated, please re-generate properties docs by running ./gradlew build-tools-docs-utils:generateConfPropsDoc -Poverwrite

```

No

### DIFF
--- a/build-tools/docs-utils/build.gradle
+++ b/build-tools/docs-utils/build.gradle
@@ -19,7 +19,9 @@
 
 ext {
     javaMainClass = "org.apache.spark.k8s.operator.utils.ConfOptionDocGenerator"
-    docsPath = System.getProperty("user.dir") + "/docs"
+    defaultDocsPath = "${project.rootDir}/docs"
+    docOutputPath = project.hasProperty("overwrite") ? defaultDocsPath : "build"
+    generatedDocFileName = "config_properties.md"
 }
 
 dependencies {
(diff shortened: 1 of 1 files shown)


jiangzho:
cc @peter-toth for review - thanks !

peter-toth:
Thanks @jiangzho, the change makes sense to me.

PR #439, opened 2025-12-20
[SPARK-54789] Support label selector based filter on resources to reconcile

jiangzho:

This PR adds a new property `spark.kubernetes.operator.reconciler.labelSelector`. If set to non empty value, operator would only reconcile custom resources that match the given selector filter. By default, operator would still reconcile all resources.

Introducing optional label based selectors allows operator to reconcile only a targeted subset of custom resources when configured explicitly.

It's a common case for user to leverage label for isolation on resources. This can be a result of various reason, including setting-up multiple operator per environment / group in shared clusters, or for progressive adoption upon new versions. This feature would make operator fits better for prod level multi-tenant environments.

New property becomes available while behavior is the same as before by default.

Existing e2e validates the default behavior. Added new scenario to test selector functionality.

No


jiangzho:
cc @peter-toth for review - many thanks in advance!

docs/config_properties.md
```diff
@@ -37,6 +37,7 @@
  | spark.kubernetes.operator.reconciler.retry.maxIntervalSeconds | Integer | -1 | false | Max interval(in seconds) of retries on unhandled controller errors. Set to non-positive for unlimited. | 
  | spark.kubernetes.operator.reconciler.terminationTimeoutSeconds | Integer | 30 | false | Grace period for operator shutdown before reconciliation threads are killed. | 
  | spark.kubernetes.operator.reconciler.trimStateTransitionHistoryEnabled | Boolean | true | true | When enabled, operator would trim state transition history when a new attempt starts, keeping previous attempt summary only. | 
+ | spark.kubernetes.operator.reconciliationSelectors | String |  | false | Comma-separated label selector(s) that the operator would be reconciling for Spark resources. If not set or set to empty string, operator would reconcile all watched Spark Apps and Clusters. |
```
dongjoon-hyun: Can we reuse the existing namespace, `spark.kubernetes.operator.reconciler`, like this? And, `selector` might be better because we use `spark-app-selector` and `spark.kubernetes.node.selector.*` before, @jiangzho .
```scala
- spark.kubernetes.operator.reconciliationSelectors
+ spark.kubernetes.operator.reconciler.selector
```

tests/e2e/resource-selector/chainsaw-test.yaml
```diff
@@ -0,0 +1,106 @@
...
+        - name: APPLICATION_FILE_NAME
+          value: spark-example.yaml
+        - name: SPARK_APPLICATION_NAME
+          value: "spark-example-resource-selector"
+        - name: SPARK_VERSION
+          value: "4.1.0"
+        - name: JAVA_VERSION
+          value: "17"
+        - name: IMAGE
+          value: "apache/spark:4.1.0-scala-java17"
```
dongjoon-hyun: Please use `apache/spark:4.1.0-scala` because we are currently advertising Java 21 over 17.

tests/e2e/resource-selector/chainsaw-test.yaml
```diff
@@ -0,0 +1,106 @@
...
+  scenarios:
+    - bindings:
+        - name: APPLICATION_FILE_NAME
+          value: spark-example.yaml
+        - name: SPARK_APPLICATION_NAME
+          value: "spark-example-resource-selector"
+        - name: SPARK_VERSION
+          value: "4.1.0"
+        - name: JAVA_VERSION
+          value: "17"
```
dongjoon-hyun: Ditto. Java 21 please.

tests/e2e/resource-selector/spark-example.yaml
```diff
@@ -0,0 +1,34 @@
...
+  jars: "local:///opt/spark/examples/jars/spark-examples.jar"
+  sparkConf:
+    spark.executor.instances: "1"
+    spark.kubernetes.container.image: ($IMAGE)
+    spark.kubernetes.authenticate.driver.serviceAccountName: "spark"
+    spark.kubernetes.driver.request.cores: "0.5"
+    spark.kubernetes.executor.request.cores: "0.5"
+  runtimeVersions:
+    sparkVersion: ($V_SPARK_VERSION)
+    jdkVersion: ($V_JAVA_VERSION)
```
dongjoon-hyun: Shall we remove this because technically `jdkVersion` didn't provide a meaningful effect in the community?

docs/config_properties.md
jiangzho: that's a good idea! I've renamed that as `spark.kubernetes.operator.reconciler.selectors` , plural just to indicate the value can include multiple selectors

tests/e2e/resource-selector/chainsaw-test.yaml
jiangzho: updated, thanks!

tests/e2e/resource-selector/spark-example.yaml
jiangzho: actually yes - we shall not care more about it

dongjoon-hyun (changes requested):
Do you have any reference for your definition of `selectors`, @jiangzho ?

> that's a good idea! I've renamed that as spark.kubernetes.operator.reconciler.selectors , plural just to indicate the value can include multiple selectors

AFAIK, in K8s, it's still a single `selector` to match all targets. For example, everything is a single selector, `fieldSelector`, `labelSelector`, `nodeSelector`, and `selector`. We don't use in a plural form, `selectors`, to specific a single set of target.

https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#resources-that-support-set-based-requirements

```
selector:
  matchLabels:
    component: redis
  matchExpressions:
    - { key: tier, operator: In, values: [cache] }
    - { key: environment, operator: NotIn, values: [dev] }
```

If you really want to use a plural form, I'd like to recommend to use `matchLabels` for the config name.

```
- spark.kubernetes.operator.reconciler.selectors
+ spark.kubernetes.operator.reconciler.matchLabels
```

tests/e2e/helm/selector-config-values.yaml
```diff
@@ -0,0 +1,18 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+operatorConfiguration:
+  spark-operator.properties: |+
+    spark.kubernetes.operator.reconciler.selectors=foo=bar
```
dongjoon-hyun: We need more test coverage to specify multiple labels.

tests/e2e/resource-selector/spark-example.yaml
```diff
@@ -0,0 +1,33 @@
...
+  namespace: default
+spec:
+  mainClass: "org.apache.spark.examples.SparkPi"
+  jars: "local:///opt/spark/examples/jars/spark-examples.jar"
+  sparkConf:
+    spark.executor.instances: "1"
+    spark.kubernetes.container.image: ($IMAGE)
+    spark.kubernetes.authenticate.driver.serviceAccountName: "spark"
+    spark.kubernetes.driver.request.cores: "0.5"
+    spark.kubernetes.executor.request.cores: "0.5"
```
dongjoon-hyun: Please remove the above two lines because those are irrelevant to this PR.

tests/e2e/helm/selector-config-values.yaml
jiangzho: ack - updated the test case such that

* Operator is configured with comma-separated key values
* Validate resources is not being reconciled if
  - no label matches the given selector string
  - resource has only one matching label, but not all
  - resource has given label, but the value does not match
* Resource is reconciled only when matching all required label selectors

dongjoon-hyun:
It seems that the following is not addressed yet. 
- https://github.com/apache/spark-kubernetes-operator/pull/439#pullrequestreview-3606314282

In addition, please rebase once more in order to resolve the conflicts.

jiangzho:
Sorry overlooked previous comment - agreed that one selector shall applied to all resources given

> Via a label selector, the client/user can identify a set of objects. The label selector is the core grouping primitive in Kubernetes.

> A label selector can be made of multiple requirements which are comma-separated. 

from https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors

I've updated the naming back to singular `labelSelector` to better indicate the nature of this feature. How does this sound ?

dongjoon-hyun:
Please update the PR description accordingly.
> This PR adds a new property spark.kubernetes.operator.reconciliationSelectors.

docs/config_properties.md
```diff
@@ -29,6 +29,7 @@
  | spark.kubernetes.operator.reconciler.clusterStatusListenerClassNames | String |  | false | Comma-separated names of SparkClusterStatusListener class implementations | 
  | spark.kubernetes.operator.reconciler.foregroundRequestTimeoutSeconds | Long | 60 | true | Timeout (in seconds) for requests made to API server. This applies only to foreground requests. | 
  | spark.kubernetes.operator.reconciler.intervalSeconds | Long | 120 | true | Interval (in seconds, non-negative) to reconcile Spark applications. Note that reconciliation is always expected to be triggered when app spec / status is updated. This interval controls the reconcile behavior of operator reconciliation even when there's no update on SparkApplication, e.g. to determine whether a hanging app needs to be proactively terminated. Thus this is recommended to set to above 2 minutes to avoid unnecessary no-op reconciliation. | 
+ | spark.kubernetes.operator.reconciler.labelSelector | String |  | false | Comma-separated label selector(s) that the operator would be reconciling for Spark resources. If not set or set to empty string, operator would reconcile all watched Spark Apps and Clusters. |
```
dongjoon-hyun: `Comma-separated label selector(s)` sounds a little weird. Could you revise the sentence.

spark-operator/src/main/java/org/apache/spark/k8s/operator/config/SparkOperatorConf.java
```diff
@@ -75,6 +75,22 @@ public final class SparkOperatorConf {
...
+  /**
+   * Comma-separated label selector(s) that the operator would be reconciling for Spark resources.
+   * If not set or set to empty string, operator would reconcile all Spark Apps and Clusters.
+   */
+  public static final ConfigOption<String> OPERATOR_RECONCILER_LABEL_SELECTOR =
+      ConfigOption.<String>builder()
+          .key("spark.kubernetes.operator.reconciler.labelSelector")
+          .enableDynamicOverride(false)
+          .description(
+              "Comma-separated label selector(s) that the operator would be reconciling for " +
```
dongjoon-hyun: ditto. We need to revise the phrase, `Comma-separated label selector(s)`.

csviri (approved):
LGTM

jiangzho:
Checked CI failures for previous commit, but seems they are not directly related to this PR

jiangzho:
cc @dongjoon-hyun - May I get another round of review for this please ? Thanks a lot

dongjoon-hyun (approved):
+1, LGTM. Thank you, @jiangzho and all.

Merged to main.

PR #446, opened 2026-01-07
[SPARK-54926] Improve Helm chart for adding labels and annotations for operator config map

jiangzho:
## What changes were proposed in this pull request?

This PR allows user to config the labels and annotations for the operator configuration `ConfigMap`, which hold operator configuration properties files.

Also added schema validation for the new fields in `values.schema.json`.

Users may need to add custom labels and annotations to the operator configuration ConfigMaps for various operational purposes, or need toolings that rely on labels or annotations on resources.

Currently, the Helm chart doesn't provide a good way to customize labels and annotations for the operator `ConfigMap` - it only support annotations for dynamic `ConfigMap`.

Users can now specify custom labels and annotations for the operator configuration ConfigMap in their `values.yaml`:

  ```yaml
  operatorConfiguration:
    configMap:
      annotations:
        example.com/annotation: "value"
      labels:
        example.com/label: "value"
```

Self-contained linters, also validated manually.

Without explicitly configured labels or annotations, the `ConfigMap` renders as

```yaml
# Source: spark-kubernetes-operator/templates/spark-operator.yaml apiVersion: v1
kind: ConfigMap
metadata:
  name: spark-kubernetes-operator-configuration
  namespace: spark-demo
  labels:
    app.kubernetes.io/name: spark-kubernetes-operator
    app.kubernetes.io/component: "operator-config"
    app.kubernetes.io/version: "0.7.0-SNAPSHOT"
    app.kubernetes.io/managed-by: Helm
    helm.sh/chart: spark-kubernetes-operator-1.5.0-dev
    spark-role: operator
data:
  log4j2.properties: |+
# ...
```

With value override

```yaml
operatorConfiguration:
  configMap:
    annotations:
      foo1: bar1
      foo2: bar2
    labels:
      foo3: bar3
```

The `ConfigMap` renders as

```yaml
# Source: spark-kubernetes-operator/templates/spark-operator.yaml apiVersion: v1
kind: ConfigMap
metadata:
  name: spark-kubernetes-operator-configuration
  namespace: spark-demo
  labels:
    app.kubernetes.io/name: spark-kubernetes-operator
    app.kubernetes.io/component: "operator-config"
    app.kubernetes.io/version: "0.7.0-SNAPSHOT"
    app.kubernetes.io/managed-by: Helm
    helm.sh/chart: spark-kubernetes-operator-1.5.0-dev
    spark-role: operator
    foo3: bar3
  annotations:
    foo1: bar1
    foo2: bar2
data:
  log4j2.properties: |+
# ...
```

No

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -177,6 +177,47 @@ jobs:
         if: matrix.mode == 'selector'
         run: |
           chainsaw test --test-dir ./tests/e2e/${{ matrix.test-group }} --parallel 1
+  helm-tests:
+    name: "Helm Tests"
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    strategy:
+      fail-fast: false
+      matrix:
+        kubernetes-version:
+          - "1.35.0"
+        test-group:
+          - configmap-metadata
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v6
+      - name: Set up JDK 25
+        uses: actions/setup-java@v4
+        with:
+          java-version: 25
+          distribution: 'zulu'
+          cache: 'gradle'
+      - name: Set up Minikube
+        uses: medyagh/setup-minikube@v0.0.21
+        with:
+          cache: true
+          kubernetes-version: ${{ matrix.kubernetes-version }}
+          cpus: 3
+          memory: 10240m
+      - name: Print K8S pods and nodes info
+        run: |
+          kubectl get pods -A
+          kubectl describe node
+      - name: Run Spark K8s Operator Helm Tests
+        run: |
+          eval $(minikube docker-env)
+          ./gradlew buildDockerImage
+          helm install spark --create-namespace -f \
+          build-tools/helm/spark-kubernetes-operator/values.yaml -f \
+          tests/e2e/helm/helm-test-values/${{ matrix.test-group }}/values.yaml \
+          build-tools/helm/spark-kubernetes-operator/
+          minikube docker-env --unset
+          helm test spark
   lint:
     name: "Linter and documentation"
     runs-on: ubuntu-latest
--- a/build-tools/helm/spark-kubernetes-operator/templates/spark-operator.yaml
+++ b/build-tools/helm/spark-kubernetes-operator/templates/spark-operator.yaml
@@ -185,7 +185,7 @@ metadata:
     {{- if index (.Values.operatorConfiguration.configMap) "labels" }}
       {{- toYaml .Values.operatorConfiguration.configMap.labels | nindent 4 }}
     {{- end }}
-  {{- if index (.Values.operatorConfiguration.configMap) "labels" }}
+  {{- if index (.Values.operatorConfiguration.configMap) "annotations" }}
   annotations:
     {{- toYaml .Values.operatorConfiguration.configMap.annotations | nindent 4 }}
   {{- end }}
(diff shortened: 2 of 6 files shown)


build-tools/helm/spark-kubernetes-operator/templates/spark-operator.yaml
```diff
@@ -210,6 +219,9 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
     {{- include "spark-operator.dynamicConfigLabels" . | nindent 4 }}
+    {{- if .Values.operatorConfiguration.dynamicConfig.labels }}
```
dongjoon-hyun: Can we spin off `dynamicConfig.labels` into a separate PR because it supports `annotations` already, @jiangzho ?

jiangzho:
Thanks @dongjoon-hyun ! Updated the commit to decouple dynamic config & added a helm test job for this

.github/workflows/build_and_test.yml
```diff
@@ -177,6 +177,47 @@ jobs:
         if: matrix.mode == 'selector'
         run: |
           chainsaw test --test-dir ./tests/e2e/${{ matrix.test-group }} --parallel 1
+  helm-tests:
+    name: "Helm Tests"
```
dongjoon-hyun, line 181: Can we extend `k8s-integration-tests` instead of adding `helm-tests` because we already have `helm test` here?

https://github.com/apache/spark-kubernetes-operator/blob/43473c7ff12c98be35a7292b00bc9dcc31665f0d/.github/workflows/build_and_test.yml#L145

tests/e2e/helm/helm-test-values/configmap-metadata/values.yaml
```diff
@@ -0,0 +1,35 @@
...
+# Usage:
+#   helm install spark-operator . -f tests/test-values-configmap-metadata.yaml
+#   helm test spark-operator
+#
+# This file demonstrates the ConfigMap labels and annotations feature
+# and can be used to validate the Helm test for this functionality.
+
+operatorConfiguration:
+  append: true
+  configMap:
```
dongjoon-hyun, line 27: I have two questions. 
- What happens when there is a conflict between the user-defined labels and annotations and the existing ones?
- Can we control extra `labels` and `annotations` for other resources of `Spark K8s Operator`?

.github/workflows/build_and_test.yml
jiangzho: It is technically feasible - this line runs helm test with the value override in this particular scenario.

This new job helps to reduce the size of `exclude` in `k8s-integration-tests ` job - even if later we add more helm tests with custom value overrides. 
If we reuse the same job, the helm tests need to exclude integration test groups & integration test modes need to exclude helm tests. This `exclude` block would have a size increase of (num of integration test groups) * (num of helm tests). That makes it less elegant an harder to maintain. Other than that I do not have concerns to merge them.

tests/e2e/helm/helm-test-values/configmap-metadata/values.yaml
jiangzho: > What happens when there is a conflict between the user-defined labels and annotations and the existing ones?

In this approach, the user-provided ones take higher precedence as they are appended to default ones.

> Can we control extra labels and annotations for other resources of Spark K8s Operator?

As of now, we support configure these for operator pod, operator & workload rbac (service accounts, roles .etc to run Spark and the operator). ConfigMap is a missing piece here.

dongjoon-hyun:
To @viirya , @peter-toth , @jiangzho , I'm going to prepare Apache Spark K8s Operator v0.7 via creating branch-0.7 in order to deliver the latest updates along with Apache Spark 4.1.x.

PR #473, opened 2026-01-29
[SPARK-55268] Fix `ConfigOption.getValue` not to invoke `resolveValue` twice

dongjoon-hyun:

This PR aims to fix `ConfigOption.getValue` method to invoke `resolveValue` only once.

Previously, `resolveValue()` is invoked twice at line 58 and 62. We can remove the second one by reusing `resolvedValue` variable.

https://github.com/apache/spark-kubernetes-operator/blob/fea548118dc4e7daf192595e20cebd8fd5bc12c3/spark-operator/src/main/java/org/apache/spark/k8s/operator/config/ConfigOption.java#L57-L63

No behavior change.

Pass the CIs.

Yes (`Opus 4.5` on `Claude Code v2.1.5`)


dongjoon-hyun:

PR #476, opened 2026-01-29
[SPARK-55274] Set `-XX:+AlwaysPreTouch` by default

dongjoon-hyun:

This PR aims to set `-XX:+AlwaysPreTouch` by default to improve stability.

To improve stability by ensuring **Physical Memory Availability**

In some environments where memory overcommitment is allowed by the OS or K8s node by SRE, the JVM might successfully reserve heap space that isn't physically available. `AlwaysPreTouch` ensures that the K8s operator actually gets the guarnteed physical RAM at startup and avoids random evictions due to the lack of memory.

Only JVM GC behavior changes.

Pass the CIs.

No.

### DIFF
--- a/build-tools/helm/spark-kubernetes-operator/values.yaml
+++ b/build-tools/helm/spark-kubernetes-operator/values.yaml
@@ -41,7 +41,7 @@ operatorDeployment:
     # https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/
     topologySpreadConstraints: [ ]
     operatorContainer:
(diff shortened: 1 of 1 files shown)


dongjoon-hyun:
Thank you, @peter-toth and @viirya . Merged to main.

melin:
Add the -XX:+UseLargePages parameter accordingly

dongjoon-hyun:
> Add the -XX:+UseLargePages parameter accordingly

It's an orthogonally independent request because it's no-op if there is no adequate OS-level setting.

However, thank you for your feedback. We can proceed that independently in another PR.

PR #484, opened 2026-02-04
[SPARK-55352] Use K8s Garbage Collection to delete executor pods

dongjoon-hyun:

This PR aims to use [K8s Garbage Collection](https://kubernetes.io/docs/concepts/architecture/garbage-collection/) to delete executor pods.

To avoid massive API invocation overhead during explicit executor deletions.

There is a timing difference for executor pod deletions. However, eventually, all executor pods are garbage collected after driver pod is deleted.

No.

### DIFF
--- a/spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppSubmissionWorker.java
+++ b/spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppSubmissionWorker.java
@@ -41,6 +41,7 @@ import org.apache.spark.deploy.k8s.submit.RMainAppResource;
 import org.apache.spark.k8s.operator.spec.ApplicationSpec;
 import org.apache.spark.k8s.operator.spec.ConfigMapSpec;
 import org.apache.spark.k8s.operator.spec.DriverServiceIngressSpec;
+import org.apache.spark.k8s.operator.spec.ResourceRetainPolicy;
 import org.apache.spark.k8s.operator.spec.RuntimeVersions;
 import org.apache.spark.k8s.operator.utils.ModelUtils;
 import org.apache.spark.k8s.operator.utils.StringUtils;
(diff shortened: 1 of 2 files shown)


spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppSubmissionWorker.java
```diff
@@ -167,6 +168,11 @@ protected SparkAppDriverConf buildDriverConf(
     effectiveSparkConf.setIfMissing("spark.app.id", appId);
     effectiveSparkConf.setIfMissing("spark.authenticate", "true");
     effectiveSparkConf.setIfMissing("spark.io.encryption.enabled", "true");
+    // Use K8s Garbage Collection instead of explicit API invocations
+    if (applicationSpec.getApplicationTolerations().getResourceRetainPolicy() !=
+        ResourceRetainPolicy.Always) {
+      effectiveSparkConf.setIfMissing("spark.kubernetes.executor.deleteOnTermination", "false");
```
viirya, line 174: So if `spark.kubernetes.executor.deleteOnTermination` is set `true` explicitly, even ResourceRetainPolicy is not `Always`, we still use `deleteOnTermination`, right?

dongjoon-hyun: Yes, the user-provided `spark.kubernetes.executor.deleteOnTermination` setting is always respected for both `true` and `false`, @viirya .

dongjoon-hyun: It's a kind of switching the default behavior only when there is no explicit setting in SparkApp YAML file.
- In Apache Spark distribution, `Driver` pod always remain. So, we need to clean up the executor pods. So, the default value of `spark.kubernetes.executor.deleteOnTermination` is true.
- In Apache Spark K8s Operator, `Driver` pod is cleaned up because `SparkApp` CRD remains. In this case, K8s is able to clean up executor pods whose owner (`Driver` pod) is gone. So, this PR aims to eliminate the explicit Spark Driver's API invocation.

viirya (approved):

dongjoon-hyun:
Thank you so much, @viirya ~

Merged to main.

PR #496, opened 2026-02-11
[SPARK-55486] Fix `StatusRecorder.patchAndStatusWithVersionLocked` not to log errors

dongjoon-hyun:

This PR aims to fix `StatusRecorder.patchAndStatusWithVersionLocked` not to log errors.

The upper layer already shows the error properly like the following. So, we had better avoid repeating the same exception log.

https://github.com/apache/spark-kubernetes-operator/blob/bdbbbbe5725e619c6bdd79ddc7bc8c11ee156c7c/spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/StatusRecorder.java#L132-L141

No behavior change because this is a removal of duplicated log.

Pass the CIs.

No.

### DIFF
--- a/spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/StatusRecorder.java
+++ b/spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/StatusRecorder.java
@@ -93,7 +93,7 @@ public class StatusRecorder<
       return;
     }
 
-    Exception err = null;
+    KubernetesClientException err = null;
     long maxRetry = API_STATUS_PATCH_MAX_ATTEMPTS.getValue();
     for (long i = 1; i <= maxRetry; i++) {
       // We retry the status update maxRetry times to avoid some intermittent connectivity errors
@@ -110,7 +110,6 @@ public class StatusRecorder<
     }
 
     if (err != null) {
-      log.error("Fail to patch status.", err);
       throw err;
     }
 


spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/StatusRecorder.java
```diff
@@ -93,7 +93,7 @@ private void patchAndStatusWithVersionLocked(CR resource, KubernetesClient clien
       return;
     }
 
-    Exception err = null;
+    KubernetesClientException err = null;
```
dongjoon-hyun, line 96: Since we use `} catch (KubernetesClientException e) {` statement, `KubernetesClientException` is the right type instead of `Exception`.

peter-toth (approved):

spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/StatusRecorder.java
csviri, line 101: Hi just a question so I understand this, why is the resource locked in this case, it seems to be if some changes the resource meanwhile, this will awlays retry up to `maxRetry` number. In this case would be benefital to re-read the resource apply the changes again and do the update. Or just don't use Optimistic locking.

In JOSDK 5.3 there will be nice utilities to make it much smarter, will come back to you if it is released.

dongjoon-hyun: To @csviri , You had better use the original PR whose author is @jiangzho .
- https://github.com/apache/spark-kubernetes-operator/pull/12

dongjoon-hyun:
Thank you, @peter-toth . Merged to main.

PR #513, opened 2026-02-19
[SPARK-55611] Add `RunningWithPartialCapacity` state for `SparkApplication` CRD

jiangzho:

This PR adds a new `RunningWithPartialCapacity` state to `ApplicationStateSummary` that indicates when a Spark application with dynamic allocation disabled is running with executor count between minimum and maximum thresholds.

When dynamic allocation is disabled, if an application has fewer executors than requested (between min and max), it indicates resource constraints in the cluster. This is a tolerable but suboptimal state that should be visible to operators.

With dynamic allocation enabled, executors scaling between min and max is normal behavior, so `RunningHealthy` is appropriate.

The new state provides better observability into cluster resource constraints affecting application performance.

Yes. Users will now see `RunningWithPartialCapacity` status for applications.

Added unit tests in `AppRunningStepTest` covering state transition with different number of executors. Also added unit tests in `ApplicationStateSummaryTest` verifying the new state summary.

No

### DIFF
--- a/spark-operator-api/src/main/java/org/apache/spark/k8s/operator/Constants.java
+++ b/spark-operator-api/src/main/java/org/apache/spark/k8s/operator/Constants.java
@@ -166,10 +166,10 @@ public class Constants {
 
   /**
    * Message indicating that the application is running with partial executor capacity (between min
-   * and max with dynamic allocation disabled).
+   * and max, exclusive, with dynamic allocation disabled).
    */
   public static final String RUNNING_WITH_PARTIAL_CAPACITY_MESSAGE =
-      "Application is running with executor count between minimum and maximum capacity.";
+      "Application is running with executor count between minimum and maximum capacity, exclusive.";
 
   /**
    * Message indicating that the application is running with less than the minimal number of
--- a/spark-operator-api/src/main/java/org/apache/spark/k8s/operator/status/ApplicationStateSummary.java
+++ b/spark-operator-api/src/main/java/org/apache/spark/k8s/operator/status/ApplicationStateSummary.java
@@ -59,7 +59,8 @@ public enum ApplicationStateSummary implements BaseStateSummary {
 
   /**
    * Application is running with dynamic allocation disabled and executor count is between minimum
-   * and maximum thresholds. The application is in a tolerable state but not at optimal capacity.
+   * and maximum thresholds, exclusive. The application is in a tolerable state but not at optimal
+   * capacity.
    */
   RunningWithPartialCapacity,
 
--- a/spark-operator/src/main/java/org/apache/spark/k8s/operator/context/SparkAppContext.java
+++ b/spark-operator/src/main/java/org/apache/spark/k8s/operator/context/SparkAppContext.java
@@ -23,7 +23,6 @@ import static org.apache.spark.k8s.operator.utils.Utils.driverLabels;
 import static org.apache.spark.k8s.operator.utils.Utils.executorLabels;
 
 import java.util.List;
-import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -144,14 +143,5 @@ public class SparkAppContext extends BaseContext<SparkApplication> {
    */
   public List<HasMetadata> getDriverResourcesSpec() {
     return getSecondaryResourceSpec().getDriverResources();
-  }
-
-  /**
-   * Returns the SparkConf from the driver configuration.
-   *
-   * @return The SparkConf object containing Spark configuration properties.
-   */
-  public Map<String, String> getSparkConf() {
-    return sparkApplication.getSpec().getSparkConf();
   }
 }
--- a/spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppRunningStep.java
+++ b/spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppRunningStep.java
@@ -23,7 +23,6 @@ import static org.apache.spark.k8s.operator.Constants.*;
 import static org.apache.spark.k8s.operator.reconciler.ReconcileProgress.completeAndDefaultRequeue;
 
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 
 import io.fabric8.kubernetes.api.model.Pod;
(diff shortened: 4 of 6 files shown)


jiangzho:
cc @peter-toth can you please help with the review ? Thanks!

peter-toth:
Looks like a nice improvement to me, although I'm not sure about the name of the new state. I wonder whether `RunningWithPartialResources` might sound better.

Also, there seems to be a UT failure, @jiangzho, can you please look into it?

cc @dongjoon-hyun as well

jiangzho:
@peter-toth would you mind re-run the failed test ? Seems a blip that caused one of the test not able to finish in 30s - which could be result of slow image pulling or similar latency. Tried to reproduce it but seems it passed at my fork (and if that's the case we may slightly increase the timeout in subsequent commits IMO)

jiangzho:
> I wonder whether RunningWithPartialResources might sound better.

As in this scenario (the app is scheduled and progressing, just with reduced parallelism) - I would consider this like a capacity shortfall, from scaling perspective. I would lean towards it comparing with "resources" as the latter is broader, and can imply not only missing executors, but also missing CPU / GPU / memory .etc. What do you think ?

spark-operator-api/src/main/java/org/apache/spark/k8s/operator/status/ApplicationStateSummary.java
```diff
@@ -57,6 +57,12 @@ public enum ApplicationStateSummary implements BaseStateSummary {
    */
   RunningHealthy,
 
+  /**
+   * Application is running with dynamic allocation disabled and executor count is between minimum
+   * and maximum thresholds. The application is in a tolerable state but not at optimal capacity.
```
dongjoon-hyun: This is correct mathematically but wrong in the general English sentence. Could you revise like the following, @jiangzho ?
```
- between minimum and maximum thresholds
+ between minimum and maximum thresholds, exclusive
```

spark-operator-api/src/main/java/org/apache/spark/k8s/operator/Constants.java
```diff
@@ -164,6 +164,13 @@ public class Constants {
   /** Message indicating that the application is running healthy. */
   public static final String RUNNING_HEALTHY_MESSAGE = "Application is running healthy.";
 
+  /**
+   * Message indicating that the application is running with partial executor capacity (between min
+   * and max with dynamic allocation disabled).
```
dongjoon-hyun: ditto.

spark-operator-api/src/main/java/org/apache/spark/k8s/operator/Constants.java
```diff
@@ -164,6 +164,13 @@ public class Constants {
...
+   */
+  public static final String RUNNING_WITH_PARTIAL_CAPACITY_MESSAGE =
+      "Application is running with executor count between minimum and maximum capacity.";
```
dongjoon-hyun: ditto.

spark-operator/src/main/java/org/apache/spark/k8s/operator/context/SparkAppContext.java
```diff
@@ -144,4 +145,13 @@ public Pod getDriverPodSpec() {
   public List<HasMetadata> getDriverResourcesSpec() {
     return getSecondaryResourceSpec().getDriverResources();
   }
+
+  /**
+   * Returns the SparkConf from the driver configuration.
+   *
+   * @return The SparkConf object containing Spark configuration properties.
+   */
+  public Map<String, String> getSparkConf() {
```
dongjoon-hyun: Could you spin-off this contribution as a new PR? In addition, please replace all existing `app.getSpec().getSparkConf()`, `sparkApplication2.getSpec().getSparkConf()`, `sparkApplication3.getSpec().getSparkConf()` with this new utility method.

spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppRunningStep.java
```diff
@@ -96,4 +111,16 @@ public ReconcileProgress reconcile(
...
   }
+
+  /**
+   * Checks if dynamic allocation is enabled for the application.
+   *
+   * @param context The SparkAppContext
+   * @return true if dynamic allocation is enabled, false otherwise
+   */
+  private boolean isDynamicAllocationEnabled(SparkAppContext context) {
+      Map<String, String> sparkConf = context.getSparkConf();
```
dongjoon-hyun: Do we need to define this variable? Let's simplify by removing this variable definition.

peter-toth:
> @peter-toth would you mind re-run the failed test ? Seems a blip that caused one of the test not able to finish in 30s - which could be result of slow image pulling or similar latency. Tried to reproduce it but seems it passed at my fork (and if that's the case we may slightly increase the timeout in subsequent commits IMO)

I tried to rerun it a few times. It is ok now.

> As in this scenario (the app is scheduled and progressing, just with reduced parallelism) - I would consider this like a capacity shortfall, from scaling perspective. I would lean towards it comparing with "resources" as the latter is broader, and can imply not only missing executors, but also missing CPU / GPU / memory .etc. What do you think ?

Yeah, I see your point. I'm ok with `PartialCapacity`.

dongjoon-hyun:
Also, @jiangzho , please add this new state into the application transition diagram.

https://github.com/apache/spark-kubernetes-operator/blob/df9d8ac848018744e373e416d3513ee19eeb2263/docs/architecture.md?plain=1#L49-L102

dongjoon-hyun (approved):
+1, LGTM (Pending CIs). Thank you for updating, @jiangzho .

dongjoon-hyun:
Oh, wait, @jiangzho . 

What happens in the following case?
- When `SparkApplication` has the same min and max values, `10`.
- After `SparkApplication` enters `RunningHealthy` with 10 executors.
- Two executor loss happens and Spark driver is trying to get two additionally, but didn't get yet.
  - In this case, the definition of `RunningWithPartialCapacity` rejects this situation because `8` is less than `min (10)` , isn't it?
  > "Application is running with executor count between minimum and maximum capacity, exclusive.";

I also added a simplified `instanceConfig` example for the case, `min=max=3`.
- #518 

https://github.com/apache/spark-kubernetes-operator/blob/04594d43b491eec2e41ea2e2f7ad509fb30ebd7b/examples/pi-with-instanceconfig.yaml#L23-L31

jiangzho:
> Two executor loss happens and Spark driver is trying to get two additionally, but didn't get yet.
> In this case, the definition of RunningWithPartialCapacity rejects this situation because 8 is less than min (10) , isn't it?

Exactly - since in this case the number of executors is less than `min` - it would fall to `RunningWithBelowThresholdExecutors`, indicating that this is not tolerable according to user configuration. `RunningWithPartialCapacity` is the state that's not fully healthy but tolerable. In this sense, I believe the given description is still accurate

jiangzho:
The failed test seems again timing out (a bit flaky to reproduce) - the 2min deadline maybe a few tight if the image is not yet cached. 

@dongjoon-hyun may we separately increase the timeout threshold a bit, like to 5min ?

dongjoon-hyun:
This one also used a wrong parent when you create a JIRA ID, @jiangzho .

Please use SPARK-55000.

Screenshot 2026-03-06 at 08 48 47

PR #514, opened 2026-02-21
[SPARK-55623] Add granular restart control with consecutive failure tracking

jiangzho:

This PR implements granular restart control for Spark applications with support for consecutive failure / scheduling failures tracking.

Consecutive failure tracking enables operators to allow more total restarts for apps with occasional transient failuresm, and to stop quickly on persistent failures. It also apply special handling for scheduling failures to mitigate API server stress.

This is an enhancement based on our current configuration for maximal restarts and backoff interval.

Yes. New optional configuration fields availale in RestartConfig:

```yaml
  restartConfig:
    restartPolicy: Always
    maxRestartAttempts: 5  # existing field
    restartBackoffMillis: 30000  # existing field
    restartCounterResetMillis: 3600000  # existing field

    # New: consecutive failure limits
    maxRestartOnFailure: 3
    restartBackoffMillisForFailure: 60000
    maxRestartOnSchedulingFailure: 2
    restartBackoffMillisForSchedulingFailure: 300000
```

This should be backwards compartible. All new fields are optional. When failure-specific limits are not set, the operator uses maxRestartAttempts as before.

Unit tests added to validate the limit evaluation flow, with & without the new fields.

No

### DIFF
--- a/spark-operator-api/src/main/java/org/apache/spark/k8s/operator/status/BaseAttemptSummary.java
+++ b/spark-operator-api/src/main/java/org/apache/spark/k8s/operator/status/BaseAttemptSummary.java
@@ -33,6 +33,6 @@ import lombok.ToString;
 @RequiredArgsConstructor
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @JsonIgnoreProperties(ignoreUnknown = true)
-public class BaseAttemptSummary<AI extends BaseAttemptInfo> {
-  protected final AI attemptInfo;
+public class BaseAttemptSummary<I extends BaseAttemptInfo> {
+  protected final I attemptInfo;
 }


jiangzho:
cc @peter-toth may I get review on this please ? Thanks a lot

docs/spark_custom_resources.md
```diff
@@ -234,14 +234,111 @@ restartConfig:
...
+| .spec.applicationTolerations.restartConfig.maxRestartAttempts                           | integer | 3             | Maximum number of restart attempts for all scenarios (always checked)                                                                                                                                                                    |
+| .spec.applicationTolerations.restartConfig.restartBackoffMillis                         | integer | 30000         | Default backoff time in milliseconds between restart attempts                                                                                                                                                                            |
+| .spec.applicationTolerations.restartConfig.maxRestartOnFailure                          | integer | null          | Maximum consecutive failures before stopping. Resets to 0 on success. If null, uses maxRestartAttempts                                                                                                                                   |
```
peter-toth, line 323: Shall we use `-1` default value for the not specified case? It seems that's the value we use with other fields as well (e.g. `restartCounterResetMillis`).

spark-operator-api/src/main/java/org/apache/spark/k8s/operator/status/AttemptInfo.java
```diff
@@ -40,13 +53,41 @@
...
-    return new AttemptInfo(id + 1L, resetRestartCounter ? 1L : restartCounter + 1);
+  public AttemptInfo createNextAttemptInfo(
+      boolean resetRestartCounter, ApplicationStateSummary currentStateSummary) {
+    long newRestartCounter = resetRestartCounter ? 1L : restartCounter + 1;
+    long newFailureCounter;
+    long newSchedulingFailureCounter;
+
+    if (resetRestartCounter) {
+      // Reset all counters when restart counter is reset
+      newRestartCounter = 1L;
```
peter-toth: This looks weird, you have already set `newRestartCounter` based on `resetRestartCounter` a few lines above.

spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppCleanUpStep.java
```diff
@@ -174,6 +174,10 @@ public ReconcileProgress reconcile(
       return appendStateAndRequeueAfter(
           context, statusRecorder, state, Duration.ofMillis(requeueAfterMillis));
     } else {
+      // Check if current state is a failure before restarting
```
peter-toth: Seems like you can move this into the if just before `requeueAfterMillis = ...`.

spark-operator-api/src/main/java/org/apache/spark/k8s/operator/status/AttemptInfo.java
peter-toth: I wonder if it would make sense to merge this `if ... else if ...` logic with the logic in `terminateOrRestart()`?
I mean, you are calculating `exceededLimit` / `stateMessage` based on `ApplicationStateSummary` there, and you are kind of calculating the new counters based on `ApplicationStateSummary` here in `AttemptInfo.createNextAttemptInfo()`. But `createNextAttemptInfo()` gets invoked only from `terminateOrRestart()`, so would it make sense to drop `createNextAttemptInfo()` entirelly, calculate everyting in `terminateOrRestart()` and keep `AttemptInfo` as a simple data class?

dongjoon-hyun:
Ack, @peter-toth . I'll take a look today.

build-tools/helm/spark-kubernetes-operator/crds/sparkclusters.spark.apache.org-v1.yaml
```diff
@@ -21996,10 +21996,14 @@ spec:
                   properties:
                     attemptInfo:
                       properties:
+                        failureRestartCounter:
+                          type: integer
```
dongjoon-hyun: Given that the PR description and code, please revert this change on `SparkCluster`. I don't think this PR has the corresponding implementation, @jiangzho .
> This PR implements granular restart control for Spark applications with support for consecutive failure / scheduling failures tracking.

docs/spark_custom_resources.md
jiangzho: I indeed thought of this - but was in split mind.

`-1` in current configuration context typically indicates `unlimited`. For example,  when `restartCounterResetMillis`, the counter is never reset. When `spark.kubernetes.operator.reconciler.retry.maxIntervalSeconds` is set to `-1`, the retry would be perform without max internal limit. Thus, I suspect setting 

```
spec.applicationTolerations.restartConfig.maxRestartAttempts: 3
spec.applicationTolerations.restartConfig.maxRestartOnFailure: -1
```

Would give user an impression that 

1. this app would be retried 3 times upon success or non-consecutive failures, and unlimited retry on consecutive failures, rather than
2. `maxRestartOnFailure` is not configured (default value -1) and hence even for consecutive failures, it should be retried 3 times at maximal.

And in fact, 1 maybe a valid requirement as well. What do you think ?

build-tools/helm/spark-kubernetes-operator/crds/sparkclusters.spark.apache.org-v1.yaml
jiangzho: ack - this was introduced as `SparkCluster` also uses `AttemptInfo`. I'll create another POJO for it without the changes

docs/spark_custom_resources.md
peter-toth: Yeah, I see your point. These configs are kind of overrides for the default `maxRestartAttempts` / `restartBackoffMillis` configs in a special case, and as such, not overriding or setting a special override value (e.g. -1 for unlimited) can be different.

Yes, maybe 1 makes sense as well.

spark-operator-api/src/main/java/org/apache/spark/k8s/operator/status/AttemptInfo.java
peter-toth: Actually, I've just tried combining the 2 logics, but it doesn't seem better than the current solution.

spark-operator-api/src/main/java/org/apache/spark/k8s/operator/spec/RestartConfig.java
```diff
@@ -42,4 +44,23 @@ public class RestartConfig {
   @Default("-1")
   @Builder.Default
   protected Long restartCounterResetMillis = -1L;
+  @Builder.Default protected Long maxRestartOnFailure = null;
```
dongjoon-hyun, line 47: ~Can we use `Integer` instead of `Long` , @jiangzho ?~

spark-operator-api/src/main/java/org/apache/spark/k8s/operator/status/ApplicationAttemptInfo.java
```diff
@@ -0,0 +1,63 @@
...
+@ToString(callSuper = true)
+@JsonInclude(JsonInclude.Include.NON_NULL)
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class ApplicationAttemptInfo extends BaseAttemptInfo {
+  protected long failureRestartCounter;
+  protected long schedulingFailureRestartCounter;
+
+  public ApplicationAttemptInfo() {
+    super();
+    failureRestartCounter = 0L;
```
dongjoon-hyun, line 53: ~For the counter, `int` is enough already, isn't it?~
```
int	32	-2,147,483,648 to 2,147,483,647
long	64	-9,223,372,036,854,775,808 to 9,223,372,036,854,775,807
```

spark-operator-api/src/main/java/org/apache/spark/k8s/operator/status/BaseAttemptSummary.java
```diff
@@ -33,6 +33,6 @@
 @RequiredArgsConstructor
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @JsonIgnoreProperties(ignoreUnknown = true)
-public class BaseAttemptSummary {
-  protected final AttemptInfo attemptInfo;
+public class BaseAttemptSummary<AI extends BaseAttemptInfo> {
+  protected final AI attemptInfo;
```
dongjoon-hyun: Oh, `AI` looks like a bad naming. Please choose something else, @jiangzho .

dongjoon-hyun (approved):
+1, LGTM (except one minor nit comment about `AI`). For the other comments, please ignore that.

spark-operator-api/src/main/java/org/apache/spark/k8s/operator/status/BaseAttemptSummary.java
jiangzho: Updated to `I` to indicate the `Info` semantic

dongjoon-hyun:
Merged to main. Thank you, @jiangzho and @peter-toth .

dongjoon-hyun:
Oh, @jiangzho . It seems that you are using a wrong parent when you create a JIRA ID. 

Screenshot 2026-03-06 at 08 47 38

Do you have any reason not to use SPARK-55000?

Screenshot 2026-03-06 at 08 48 47

PR #530, opened 2026-02-28
[SPARK-55750] Retry creating secondary resources on `408/500/502/503/504` error codes and network timeouts

jiangzho:

This commit extends the existing retry logic in `ReconcilerUtils::getOrCreateSecondaryResource` to also retry on transient API server errors (408, 502, 503, 504), 500, and network-level timeouts, besides 409 conflict.

Also added unit test coverage for the retry logic.

Transient API server overload and network-level timeouts are recoverable conditions. Without retrying, a single dropped response causes the entire reconcile loop to fail and would mark app as scheduling failed. This enhance the scheduling logic robustness.

No behavioral change for healthy apps. On transient API server failures, the operator now retries up to `spark.kubernetes.operator.api.secondaryResourceCreateMaxAttempts` times (default 3) rather than immediately propagating the error.

Added unit test to validate the retry behavior.

No

### DIFF
--- a/spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/ReconcilerUtils.java
+++ b/spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/ReconcilerUtils.java
@@ -225,7 +225,7 @@ public final class ReconcilerUtils {
   private static boolean isTransientError(KubernetesClientException e) {
     // code 0 is fabric8's sentinel for network-level failures (timeouts, connection resets, etc.)
     return switch (e.getCode()) {
-      case 0, HTTP_CLIENT_TIMEOUT, HTTP_INTERNAL_ERROR, HTTP_BAD_GATEWAY,
+      case 0, HTTP_CLIENT_TIMEOUT, HTTP_BAD_GATEWAY,
            HTTP_UNAVAILABLE, HTTP_GATEWAY_TIMEOUT -> true;
       default -> false;
     };


jiangzho:
cc @peter-toth for review - thanks a lot!

peter-toth:
> cc @peter-toth for review - thanks a lot!

I will try to check this tomorrow or Thursday.

peter-toth:
cc @dongjoon-hyun

dongjoon-hyun (changes requested):
Please enumerate all error codes which this PR handles additionally.

Otherwise, the title `transient 5xx` might be misleading (even it includes `transient` explicitly)

spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/ReconcilerUtils.java
```diff
@@ -211,6 +221,16 @@ public static <T extends HasMetadata> void deleteResourceIfExists(
     }
   }
 
+  private static boolean isTransientError(KubernetesClientException e) {
+    // code 0 is fabric8's sentinel for network-level failures (timeouts, connection resets, etc.)
+    int code = e.getCode();
+    return code == 0
+        || code == HTTP_INTERNAL_ERROR
```
dongjoon-hyun: Well, this is a little tricky.

408 (Request Timeout), 429 (Too Many Requests), 502 (Bad Gateway), 503 (Service Unavailable), 504 (Gateway Timeout) are transient, but 500 (Internal Server Error) is case-by-case.

If we want to keep `isTransientError` method name, we should move `500` out from this method.

dongjoon-hyun: To be clear, I agree that we need to retry on 500 (Internal Server Error) error case, but it's not always categorized into a transient error.

spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/ReconcilerUtils.java
```diff
@@ -211,6 +221,16 @@ public static <T extends HasMetadata> void deleteResourceIfExists(
...
+        || code == HTTP_BAD_GATEWAY
+        || code == HTTP_UNAVAILABLE
+        || code == HTTP_GATEWAY_TIMEOUT;
```
dongjoon-hyun: We had better use `switch` statement in this case. Something like the following:
```java
return switch (e.getCode()) {
  case 0, HTTP_INTERNAL_ERROR, HTTP_BAD_GATEWAY,
      HTTP_UNAVAILABLE, HTTP_GATEWAY_TIMEOUT -> true;
  default -> false;
};
```

spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/ReconcilerUtils.java
jiangzho: Thanks for the catch! Yes, 500 should not be defined as transient. I've updated the naming (and also included 408).

For 429, I'll try to handle that with separate JIRA - as server is rate limiting it's better to be retried with backoff.

spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/ReconcilerUtils.java
```diff
@@ -211,6 +222,15 @@ public static <T extends HasMetadata> void deleteResourceIfExists(
     }
   }
 
+  private static boolean isTransientError(KubernetesClientException e) {
+    // code 0 is fabric8's sentinel for network-level failures (timeouts, connection resets, etc.)
+    return switch (e.getCode()) {
+      case 0, HTTP_CLIENT_TIMEOUT, HTTP_INTERNAL_ERROR, HTTP_BAD_GATEWAY,
```
dongjoon-hyun: `HTTP_INTERNAL_ERROR` should not be in `isTransientError`, @jiangzho .

Maybe, this is a simple copy-and-paste of my style change example?

dongjoon-hyun:
Merged to main

PR #641, opened 2026-04-22
[SPARK-56115] Support `crds.install` in Helm Chart

TQJADE:
This PR introduces a crds.install configuration option to the Helm chart. CRD files are moved from the crds/ directory
(unconditionally installed by Helm) to templates/crds/ with a {{- if .Values.crds.install }} guard, allowing users to
disable CRD installation by setting crds.install: false. Similar to:
https://github.com/argoproj/argo-helm/blob/f7aee451eb788abd682d187fb055668bbfe28091/charts/argo-cd/values.yaml#L31-L40

In environments where CRDs are managed separately, users need the ability to skip CRD installation at the chart level. Previously the only option was the helm install --skip-crds flag, which is not declarative and cannot be expressed in a values file.

Yes. 

* crds.install (default: true) — Controls whether CRDs are installed with the chart. Set to false when CRDs are managed  externally or when deploying a second operator instance.
* crds.annotations (default: {"helm.sh/resource-policy": "keep"}) — Annotations applied to CRD resources. The default prevents CRDs from being deleted on helm uninstall.                                                                       
* Migration note: Users running multiple operator releases in different namespaces must set crds.install: false on 
releases.

```
helm template --set crds.install=true renders both CRDs
helm template --set crds.install=false renders no CRDs                                                                    
```
Generated-by: Claude Opus 4.7

### DIFF
--- a/spark-operator-api/build.gradle
+++ b/spark-operator-api/build.gradle
@@ -68,7 +68,7 @@ tasks.register("assertGeneratedCRDMatchesHelmChart") {
   def currentPath = projectDir.absolutePath
   doLast {
     def generatedCRDFileBase = "$currentPath/build/resources/main/"
-    def stagedCRDFileBase = "$currentPath/../build-tools/helm/spark-kubernetes-operator/crds/"
+    def stagedCRDFileBase = "$currentPath/../build-tools/helm/spark-kubernetes-operator/templates/crds/"
     def generatedAppCRD = [
         "yq",
         "e",
@@ -82,16 +82,12 @@ tasks.register("assertGeneratedCRDMatchesHelmChart") {
         "${generatedCRDFileBase}sparkclusters.spark.apache.org-v1.yml"
     ].execute().text.trim()
     def stagedAppCRD = [
-        "yq",
-        "e",
-        ".spec.versions[-1]",
-        "${stagedCRDFileBase}sparkapplications.spark.apache.org-v1.yaml"
+        "bash", "-c",
+        "sed '/^{{/d' ${stagedCRDFileBase}sparkapplications.spark.apache.org-v1.yaml | yq e '.spec.versions[-1]' -"
     ].execute().text.trim()
     def stagedClusterCRD = [
-        "yq",
-        "e",
-        ".spec.versions[-1]",
-        "${stagedCRDFileBase}sparkclusters.spark.apache.org-v1.yaml"
+        "bash", "-c",
+        "sed '/^{{/d' ${stagedCRDFileBase}sparkclusters.spark.apache.org-v1.yaml | yq e '.spec.versions[-1]' -"
     ].execute().text.trim()
     if (generatedAppCRD != stagedAppCRD || generatedClusterCRD != stagedClusterCRD) {
       def errorMessage = new StringBuilder("Generated CRD yaml does not match the staged " +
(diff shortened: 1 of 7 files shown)


spark-operator-api/build.gradle
```diff
@@ -68,7 +68,7 @@ tasks.register("assertGeneratedCRDMatchesHelmChart") {
   def currentPath = projectDir.absolutePath
   doLast {
     def generatedCRDFileBase = "$currentPath/build/resources/main/"
-    def stagedCRDFileBase = "$currentPath/../build-tools/helm/spark-kubernetes-operator/crds/"
+    def stagedCRDFileBase = "$currentPath/../build-tools/helm/spark-kubernetes-operator/templates/crds/"
```
dongjoon-hyun, line 71: Is this safe?

dongjoon-hyun (changes requested):
I'm not sure if this PR is safe because `templates/crds` seems to be not protected during uninstallation. Could you double-check the CRD remains in the system during `helm uninstall` operation like before?

> ### Does this PR introduce any user-facing change?
> No

dongjoon-hyun:
Why do we need to change `spark-operator-api/build.gradle` in this PR which adds `crds.install`?

TQJADE:
> I'm not sure if this PR is safe because `templates/crds` seems to be not protected during uninstallation. Could you double-check the CRD remains in the system during `helm uninstall` operation like before?
> 
> > ### Does this PR introduce any user-facing change?
> > No

Thanks for catching this! The move from crds/ to templates/crds/ did lose Helm's built-in uninstall protection. I've addressed this by adding a configurable crds.annotations value, defaulting to "helm.sh/resource-policy": keep, which tells Helm to skip deleting CRDs on uninstall.

TQJADE:
> Why do we need to change `spark-operator-api/build.gradle` in this PR which adds `crds.install`?

1. The compared base file path change: The CRD files were moved from crds/ to templates/crds/
2. Crd contents contains the `{{- if .Values.crds.install }}` etc, `yq` cmd will failed with `sed`

dongjoon-hyun:
Could you check the CI failures because it fails 3 times consistently in our CI? Does it pass locally?

Screenshot 2026-04-24 at 15 40 59

TQJADE:
> Could you check the CI failures because it fails 3 times consistently in our CI? Does it pass locally?
> 
> Screenshot 2026-04-24 at 15 40 59

Fix by commit: 4bfbb74ab9bc5ea8e52c2799df4695334af01875

dongjoon-hyun (commented):
I'm wondering if you verify `helm upgrade` manually, too?

TQJADE:
> I'm wondering if you verify `helm upgrade` manually, too?

Thanks for pointing this out. Yes, upgrade will failed without labeling and annotations. Adding doc: https://github.com/apache/spark-kubernetes-operator/pull/641/changes/382597984e082296ab55182b246e4da84f0602d3#diff-ad74991b96593fdee570962fc5b1e319080abd55e16614f556f74b7de8fd02d1R161

dongjoon-hyun (changes requested):
Hi, @TQJADE . Thank you for all your efforts and passion.

Given the current context, the potential side effects of this PR seem quite complex. Since we haven't had sufficient time to fully verify them, moving forward right now feels a bit risky. There's a concern that this might inadvertently lock us into supporting custom CRDs indefinitely, whereas the Apache Spark community is aiming for a more homogeneous infrastructure based on standard, community-blessed CRDs.

Therefore, I'd suggest we focus on releasing v1.0 without this feature for now, and then revisit this topic carefully for v2.0 once we've had time for more comprehensive testing.

PR #692, opened 2026-05-27
[SPARK-57106] Disallow watched namespace changes via dynamic config

dongjoon-hyun:

Stop the dynamic-config reconciler from changing the operator's watched namespaces. The `namespaceUpdater` wiring is removed from `SparkOperatorConfigMapReconciler`, and the now-orphaned `updateWatchingNamespaces` method (and `registeredSparkControllers` field) are removed from `SparkOperator`. `watchedNamespaces` is set once in the constructor from Helm values.

A write to the dynamic-config ConfigMap could expand the operator's watched namespaces at runtime, widening its blast radius without a restart or RBAC change. Watched-namespace scope should be defined at deploy time only.

Yes. Changes to the watched-namespaces key in the dynamic-config ConfigMap are now ignored. Operators must be restarted with updated Helm values to change watched namespaces. Other dynamic-config keys are unchanged.

Updated `SparkOperatorConfigMapReconcilerTest` and `SparkOperatorTest` (removed the obsolete namespace-update test). Verified with `gradle :spark-operator:test`.

Generated-by: Claude Code (Opus 4.7)


peter-toth (approved):
LGTM, but do we need a config to trigger legacy behaviour or was it always unsafe?

viirya (approved):

dongjoon-hyun:
It's always unsafe and the user needs to reinstall via Helm.
>  Operators must be restarted with updated Helm values to change watched namespaces.

dongjoon-hyun:
Thank you, @peter-toth and @viirya .

aaruna:
> It's always unsafe and the user needs to reinstall via Helm.

@dongjoon-hyun Pls clarify what's unsafe. 
The Spark operator admins control access to the spark operator namespace and so control the dynamic config. access.  So as long as the operator is only looking at the configmap within its namespace, how is this a concern? Is there some additional context?

dongjoon-hyun (commented):
There are some requests to keep the previous features because it's protected safely in the properly configured environments. I'm holding on this PR and trying to keep the existing features in a way.

dongjoon-hyun:
Thank you all. I close this PR in favor of the following.
- https://github.com/apache/spark-kubernetes-operator/pull/707

PR #699, opened 2026-05-30
[SPARK-57157] Harden fork CI status workflows against run/check-run race and pagination

viirya:

Port the hardening fixes from Apache Spark (SPARK-57154, SPARK-57155) to this
repository's `notify_test_workflow.yml` and `update_build_status.yml`, which were
introduced here as a copy of Spark's fork-based CI status mechanism (SPARK-57153).

`notify_test_workflow.yml`:

1. When listing the fork's workflow runs, instead of blindly taking the most
   recent run (`workflow_runs[0]`) and throwing if its `head_sha` does not match
   the PR head SHA, retry (up to 3 times, 3s apart) looking for the run whose
   `head_sha` matches the PR head SHA. The listing endpoint orders by most recent,
   so the run for the just-pushed SHA may not be registered yet and a stale run
   from a previous push could be returned.

2. When resolving the `Run / License Check` check-run id (used only to render a
   Check-run view link instead of the Actions view, see SPARK-37879), a missing
   check-run no longer throws. The check-run materializes later than the workflow
   run, especially when the matrix is queued, so this is now best-effort: if it
   cannot be found, the `Build` check is still created pointing at the Actions
   run URL.

`update_build_status.yml`:

3. List a commit's check-runs with `github.paginate(..., per_page: 100)` instead
   of a single un-paginated request, matching `notify_test_workflow.yml`. The
   default page size is 30, so the target `Build` check could fall off the first
   page on a SHA that accumulates more check-runs than that.

4. Wrap `JSON.parse(cr.output.text)` in try/catch and `continue` on failure, so a
   `Build` check with empty or malformed output text does not abort the whole
   scheduled run and block updates for every PR queued behind it.

The race conditions previously left a PR with no `Build` check at all, and the
scheduled updater only syncs existing checks, so the PR had no status reported
until the next push. The pagination and parsing issues silently block status
updates, leaving PRs stuck in `queued`.

No. CI infrastructure only.

Static verification: the embedded `actions/github-script` bodies pass
`node --check`, and the workflow YAML parses.

### DIFF
--- a/.github/workflows/notify_test_workflow.yml
+++ b/.github/workflows/notify_test_workflow.yml
@@ -61,22 +61,48 @@ jobs:
             console.log('Ref: ' + context.payload.pull_request.head.ref)
             console.log('SHA: ' + context.payload.pull_request.head.sha)
 
-            // Wait 3 seconds to make sure the fork repository triggered a workflow.
-            await new Promise(r => setTimeout(r, 3000))
-
-            let runs
-            try {
-              runs = await github.request(endpoint, params)
-            } catch (error) {
-              console.error(error)
-              // Assume that runs were not found.
-            }
-
             const name = 'Build'
             const head_sha = context.payload.pull_request.head.sha
             let status = 'queued'
 
-            if (!runs || runs.data.workflow_runs.length === 0) {
+            // Wait 3 seconds to make sure the fork repository triggered a workflow.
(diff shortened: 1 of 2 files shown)


dongjoon-hyun (approved):
+1, LGTM. Thank you, @viirya .

viirya:
Thank you @dongjoon-hyun

PR #702, opened 2026-06-09
[SPARK-57350] Enable `spark.kubernetes.executor.useDriverPodIP` by default

dongjoon-hyun:

This PR sets `spark.kubernetes.executor.useDriverPodIP` to `true` by default via
`setIfMissing` in `SparkAppSubmissionWorker#buildDriverConf`, next to the other
operator-injected defaults. Since it uses `setIfMissing`, an explicit value in the
application spec is always respected.

`spark.kubernetes.executor.useDriverPodIP` (added in Spark 4.1.0 via SPARK-53944) makes
executor pods connect to the driver using the driver pod IP instead of the driver
`Service` hostname, bypassing Kubernetes DNS. The operator fully manages the driver pod
lifecycle, and in cluster mode the driver is not restarted (the application terminates if
it dies), so the pod IP is stable for the application's lifetime. Enabling it by default
avoids DNS propagation delays/failures on executor startup and reduces CoreDNS load. It is
a no-op for Spark versions older than 4.1.0.
- https://github.com/apache/spark/pull/52650

Yes. When `spark.kubernetes.executor.useDriverPodIP` is unset, the operator now defaults
it to `true` (previously Spark's default of `false`). Applications that set it explicitly
are unaffected.

Pass the CIs.

### DIFF
--- a/spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppSubmissionWorker.java
+++ b/spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppSubmissionWorker.java
@@ -168,6 +168,7 @@ public class SparkAppSubmissionWorker {
     effectiveSparkConf.setIfMissing("spark.authenticate", "true");
     effectiveSparkConf.setIfMissing("spark.io.encryption.enabled", "true");
     effectiveSparkConf.setIfMissing("spark.kubernetes.driver.annotateExitException", "true");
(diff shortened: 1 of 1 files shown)


peter-toth (approved):

dongjoon-hyun:
Thank you, @peter-toth !

viirya (approved):

dongjoon-hyun:
Thank you, @viirya !

dongjoon-hyun:
Merged to main

PR #707, opened 2026-06-15
[SPARK-57466] Support `s.k.o.dynamicConfig.source` to use mounted-`ConfigMap`

csviri:
  
  This PR adds a second, mount-based source for dynamic configuration overrides and makes the dynamic-config mechanism pluggable, while retaining the existing ConfigMap-informer behavior for backwards compatibility.

  A new config option `spark.kubernetes.operator.dynamicConfig.source` selects the source:

  - **`configMap`** (default) — the existing behavior: a dedicated `Operator` runs `SparkOperatorConfigMapReconciler`, watching a ConfigMap via a Kubernetes informer. Requires RBAC to read ConfigMaps.
  - **`file`** — a new `DynamicConfigMonitor` periodically reloads a properties file mounted from a ConfigMap volume, requiring no extra RBAC.

  Both paths funnel through the same `SparkOperatorConfManager.refresh(...)` + watched-namespace update logic, so runtime behavior is identical once overrides are applied.

  Key changes:
  
  - **New `DynamicConfigMonitor`** (`config/DynamicConfigMonitor.java`) — a single-thread scheduled reader of a properties file with `start()` / `stop()` / `isRunning()` lifecycle and change detection against the last loaded
  snapshot.
  - **`SparkOperator`** — `registerDynamicConfig()` dispatches on `dynamicConfig.source`: the `file` source returns a `DynamicConfigMonitor`; the `configMap` source registers the second informer `Operator` (restored
  `registerSparkOperatorConfMonitor()` / `overrideConfigMonitorConfigs()`). The operator still tracks a `List<Operator>` plus an optional monitor.
  - **`SparkOperatorConf`** — added `dynamicConfig.source`, `dynamicConfig.filePath`, `dynamicConfig.reloadIntervalSeconds`; retained `dynamicConfig.selector` and `dynamicConfig.reconcilerParallelism`.
  - **Probes** (`HealthProbe`, `ReadinessProbe`, `ProbeService`, `ProbeUtil`) — health/readiness cover all registered operators (so the informer operator is checked under `configMap`) and additionally the monitor's running
  state under `file`.
  - **Helm chart**:
    - `dynamicConfig.source` added to `values.yaml` (default `configMap`).
    - The ConfigMap `Role`/`RoleBinding` is created only when `enable && source == configMap`.
    - The dynamic-config volume and mount are added only when `enable && source == file`.
    - The dynamic ConfigMap renders a single `spark-operator-dynamic.properties` key for the `file` source, or raw per-key data for the `configMap` source.

  The informer-based source requires the operator to hold cluster/namespace RBAC on ConfigMaps and runs a second JOSDK `Operator` purely to watch one ConfigMap. The mount-based source removes that RBAC requirement and the extra
  informer by letting the kubelet project ConfigMap changes onto a volume that the operator polls. Rather than forcing a migration, this PR offers the mount-based approach as an opt-in alternative while keeping the informer as
  the default, so existing deployments are unaffected.
  
  Yes, additive and backwards-compatible:

  - New `spark.kubernetes.operator.dynamicConfig.source` option (default `configMap`). With the default, behavior is unchanged from the released operator.
  - New `dynamicConfig.filePath` and `dynamicConfig.reloadIntervalSeconds` options, used only by the `file` source.
  - New `dynamicConfig.source` Helm value (default `configMap`). Opting into `file` switches the chart to mount the ConfigMap as a volume (rendered under a single `spark-operator-dynamic.properties` key) and drops the ConfigMap
  RBAC `Role`/`RoleBinding`.

  No change for users who do not set `dynamicConfig.source`.

  - Unit tests for both sources:
    - `configMap` — `SparkOperatorConfigMapReconcilerTest` and the configMap-source case in `SparkOperatorTest` (asserts a second `Operator` is registered and no `DynamicConfigMonitor`).
    - `file` — `DynamicConfigMonitorTest` (initial load, no-op reload when unchanged, refresh + namespace update on change, `isRunning()` lifecycle) and the file-source case in `SparkOperatorTest`.
  - Updated `HealthProbeTest`, `ReadinessProbeTest`, `ProbeServiceTest` for the `List<Operator>` + monitor signatures.
  - e2e `watched-namespaces` chainsaw test runs against the `file` source (`tests/e2e/helm/dynamic-config-values{,-2}.yaml` set `source: file`); ConfigMaps use the `spark-operator-dynamic.properties` layout and the propagation
  wait was raised to account for kubelet volume-sync plus the reload interval.


csviri:
this is a strange failure, tests are passing:
https://github.com/apache/spark-kubernetes-operator/actions/runs/26576971235

dongjoon-hyun:
Thank you, @csviri .

For the following question, it's just time-out due to the limit of `ubuntu-slim` GitHub Actions Virtual Environment. Please re-trigger the failed pipeline. We use `ubuntu-slim` because of the cost-savings.

> this is a strange failure, tests are passing: https://github.com/apache/spark-kubernetes-operator/actions/runs/26576971235

Screenshot 2026-06-15 at 09 28 17

docs/config_properties.md
```diff
@@ -10,9 +10,10 @@
...
+ | spark.kubernetes.operator.dynamicConfig.enabled | Boolean | false | false | When enabled, operator periodically reloads property overrides from a properties file mounted from a ConfigMap. The file path is controlled by spark.kubernetes.operator.dynamicConfig.filePath. | 
+ | spark.kubernetes.operator.dynamicConfig.filePath | String | /opt/spark-operator/dynamic-conf/spark-operator-dynamic.properties | false | Path of the properties file holding dynamic configuration overrides. Typically populated by mounting a ConfigMap as a volume. | 
  | spark.kubernetes.operator.dynamicConfig.reconcilerParallelism | Integer | 1 | false | Parallelism for dynamic config reconciler. Unbounded pool would be used if set to non-positive number. | 
- | spark.kubernetes.operator.dynamicConfig.selector | String | app.kubernetes.io/name=spark-kubernetes-operator,app.kubernetes.io/component=operator-dynamic-config-overrides | false | The selector str applied to dynamic config map. |
```
dongjoon-hyun, line 15: BTW, @csviri . This is intrusive. In general, Apache Spark community tries to keep old features as legacy alternatives. Please recover the old feature? I also didn't merge my original work due to that.

csviri: @dongjoon-hyun will do, thank you!

csviri:
@dongjoon-hyun updated, please take a look if this way with feature flag is how you imagined this. thx!

dongjoon-hyun:
Thank you, @csviri .

cc @aaruna , too.

jiangzho (approved):
Love this feature - thanks @csviri !

spark-operator/src/test/java/org/apache/spark/k8s/operator/probe/HealthProbeTest.java
```diff
@@ -86,48 +78,12 @@ public boolean isStarted() {
...
-            return unhealthyEventSources2;
-          }
-        };
-
     when(operator.getRuntimeInfo()).thenReturn(runtimeInfo);
-    when(sparkConfMonitor.getRuntimeInfo()).thenReturn(runtimeInfo2);
-  }
-
-  @Test
-  void testHealthProbeWithInformerHealthWithMultiOperators() {
```
dongjoon-hyun, line 109: May I ask if this is inevitable removal of irrelevant `HealthProbe` test coverage, @csviri ?

dongjoon-hyun: The new `testHealthProbeWithDynamicConfigMonitor` test coverage contains this?

tests/e2e/watched-namespaces/spark-operator-dynamic-config-1.yaml
```diff
@@ -17,7 +17,8 @@
 
 apiVersion: v1
 data:
-  spark.kubernetes.operator.watchedNamespaces: default, spark-1
+  spark-operator-dynamic.properties: |
+    spark.kubernetes.operator.watchedNamespaces=default, spark-1
```
dongjoon-hyun: Do we need to change this?

tests/e2e/watched-namespaces/spark-operator-dynamic-config-2.yaml
```diff
@@ -17,7 +17,8 @@
 
 apiVersion: v1
 data:
-  spark.kubernetes.operator.watchedNamespaces: default
+  spark-operator-dynamic.properties: |
+    spark.kubernetes.operator.watchedNamespaces=default
```
dongjoon-hyun: Do we need to change this?

tests/e2e/watched-namespaces/spark-operator-dynamic-config-1.yaml
csviri: yes, see below.

tests/e2e/watched-namespaces/spark-operator-dynamic-config-2.yaml
csviri: Same as above: yes, afaik there is no way to mount all ConfigMap data as a single file in K8S. A key is always a new file. So we have to place the whole content under a key in ConfigMap.

spark-operator/src/test/java/org/apache/spark/k8s/operator/probe/HealthProbeTest.java
csviri: No, nit properly, good catch, fixed that and `ProbeServiceTest`. thx!

dongjoon-hyun:
Hi, @csviri . According to your comment, the following sounds irrelevant to this PR's proposal. Could you make a new PR only for that? We can proceed that independent topic before this PR.
> yes, afaik there is not way to mount all ConfigMap data in K8S. A key is always a new file.

```yaml
data:
- spark.kubernetes.operator.watchedNamespaces: default
+   spark-operator-dynamic.properties: |
+        spark.kubernetes.operator.watchedNamespaces=default
```

csviri:
> Hi, @csviri . According to your comment, the following sounds irrelevant to this PR's proposal. Could you make a new PR only for that? We can proceed that independent topic before this PR.
> 
> > yes, afaik there is not way to mount all ConfigMap data in K8S. A key is always a new file.
> 
> ```yaml
> data:
> - spark.kubernetes.operator.watchedNamespaces: default
> +   spark-operator-dynamic.properties: |
> +        spark.kubernetes.operator.watchedNamespaces=default
> ```

@dongjoon-hyun maybe I was not clear, those are needed/used in the chainsaw test, and needs to be in this format. Or do I miss something?

dongjoon-hyun:
No, you were clear. 
> maybe I was not clear, those are needed/used in the chainsaw test, and needs to be in this format. Or do I miss something?

Maybe, I wasn't clear enough. My perspective was the following.
- This is supposed to introduce a new alternative path which we should not affect the existing functionality and **test coverage**.
- In other words, this PR is supposed to introduce an independent set of test coverage instead of touching the existing test code and test resource.

So, if you need to change some existing test resource for any reasons like the following, it should be done before this PR independently.
> A key is always a new file.

csviri:
> No, you were clear.
> 
> > maybe I was not clear, those are needed/used in the chainsaw test, and needs to be in this format. Or do I miss something?
> 
> Maybe, I wasn't clear enough. My perspective was the following.
> 
> * This is supposed to introduce a new alternative path which we should not affect the existing functionality and **test coverage**.
> * In other words, this PR is supposed to introduce an independent set of test coverage instead of touching the existing test code and test resource.
> 
> So, if you need to change some existing test resource for any reasons like the following, it should be done before this PR independently.
> 
> > A key is always a new file.

Ahh I see, change the PR! thx!

spark-operator/src/main/java/org/apache/spark/k8s/operator/config/DynamicConfigMonitor.java
```diff
@@ -0,0 +1,160 @@
...
+   * running.
+   */
+  public boolean isRunning() {
+    return initialLoadComplete.get() && !scheduler.isShutdown();
+  }
+
+  private void reloadSafely() {
+    try {
+      reload();
+    } catch (UncheckedIOException e) {
```
dongjoon-hyun: This looks unsafe to me. For example,
- `Properties.load` can throw `IllegalArgumentException` which is not `UncheckedIOException`. 
- `namespaceUpdater.accept(watchedNamespacesSupplier.get())` can throw `RuntimeException` which is not `UncheckedIOException`.

Please double-check the code patch and make this robust.

spark-operator/src/main/java/org/apache/spark/k8s/operator/config/DynamicConfigMonitor.java
```diff
@@ -0,0 +1,160 @@
...
+  private void reload() {
+    Map<String, String> current = readProperties();
+    if (current.equals(lastLoaded.get())) {
+      return;
+    }
+    log.info(
+        "Detected dynamic config change in {}, applying {} overrides", configFile, current.size());
+    SparkOperatorConfManager.INSTANCE.refresh(current);
+    lastLoaded.set(current);
+    namespaceUpdater.accept(watchedNamespacesSupplier.get());
```
dongjoon-hyun, line 145: This seems to break `reload`'s assumption. This means we will not retry when `namespaceUpdater.accept(watchedNamespacesSupplier.get())` fails, doesn't it?

spark-operator/src/main/java/org/apache/spark/k8s/operator/config/SparkOperatorConf.java
```diff
@@ -235,32 +235,71 @@ public final class SparkOperatorConf {
...
+              "Path of the properties file holding dynamic configuration overrides. Used by the "
+                  + "'file' source, typically populated by mounting a ConfigMap as a volume.")
+          .typeParameterClass(String.class)
+          .defaultValue("/opt/spark-operator/dynamic-conf/spark-operator-dynamic.properties")
+          .build();
+
+  /** Interval at which the dynamic config file is re-read (used by the {@code file} source). */
+  public static final ConfigOption<Long> DYNAMIC_CONFIG_RELOAD_INTERVAL_SECONDS =
+      ConfigOption.<Long>builder()
+          .key("spark.kubernetes.operator.dynamicConfig.reloadIntervalSeconds")
```
dongjoon-hyun, line 294: We need to accept only positive value. For 0 or negative, `operator` will fail due to `IllegalArgumentException` at start-up.

spark-operator/src/main/java/org/apache/spark/k8s/operator/probe/HealthProbe.java
```diff
@@ -60,7 +62,6 @@ public boolean isHealthy() {
     if (!areOperatorsStarted(operators).orElse(false)) {
       return false;
     }
-
```
dongjoon-hyun, line 63: nit. Please don't mix a style-change into this PR. Simply, recover the original empty line and reduce the footprint (size) of this PR.

dongjoon-hyun (changes requested):
Thank you so much. It's much better. I left a few more comments on the corner cases, @csviri .

spark-operator/src/main/java/org/apache/spark/k8s/operator/probe/HealthProbe.java
csviri: fixed, thx!

spark-operator/src/main/java/org/apache/spark/k8s/operator/config/SparkOperatorConf.java
csviri: I added check! thx!

spark-operator/src/main/java/org/apache/spark/k8s/operator/config/DynamicConfigMonitor.java
csviri: Now we are catching all RuntimeException. WDYT? thx!

spark-operator/src/main/java/org/apache/spark/k8s/operator/config/DynamicConfigMonitor.java
csviri: I altered the implementation, now it has more consistent behavior in that regard with the controller based implementation. Please let me know, if it is good enough this way. thx!

dongjoon-hyun (approved):
+1, LGTM. Thank you so much, @csviri .

dongjoon-hyun:
Merged to main

PR #708, opened 2026-06-16
[SPARK-57492] Use K8s Garbage Collection to delete executors only when restart is disabled

dongjoon-hyun:

This PR applies the `spark.kubernetes.executor.deleteOnTermination=false` override from
SPARK-55352 only when the application does not restart, i.e. `RestartPolicy` is `Never` or
unset, in addition to the existing static-allocation condition.

When `RestartPolicy=Always` (like `spark-thrift-server.yaml`), the driver pod can be recreated while old executor pods linger, so they accumulate instead of being garbage collected. Such applications must keep the default behavior which delete executors on termination explicitly.

https://github.com/apache/spark-kubernetes-operator/blob/e875263375e5cd5bac7e160d8d7b913b9b9c28e6/examples/spark-thrift-server.yaml#L34

No.

Pass the CIs with the updated `useGarbageCollectionToDeleteExecutors` unit test.


peter-toth (approved):

dongjoon-hyun:
Thank you, @peter-toth !

dongjoon-hyun:
Merged to main

PR #710, opened 2026-06-16
[SPARK-57495] Default `driver/master/worker` container `allowPrivilegeEscalation` to false if unset

dongjoon-hyun:

Add a `SecurityContextDecorator` that sets each container's `securityContext.allowPrivilegeEscalation` to `false` when it is unset, and apply it to the operator-created Pods:

- driver Pod (`SparkApplication`) in `SparkAppResourceSpecFactory`
- master/worker `StatefulSet`s (`SparkCluster`) in `SparkClusterResourceSpecFactory`

It only fills in the value when null, so a value the user set explicitly is preserved. Executor Pods (created by the Spark driver, not the operator) are out of scope.

An unset `allowPrivilegeEscalation` is treated as `true` by Kubernetes, which violates the Pod Security Standards "restricted" profile. Defaulting it to `false` hardens operator-created Pods by default.

In addition, this is consistent with Apache Spark 4.3.0 behavior.
- https://github.com/apache/spark/pull/56549

Yes. Driver/master/worker containers now get `securityContext.allowPrivilegeEscalation: false` when the user did not set it. An explicitly set value (e.g. `true`) is kept.

Added `SecurityContextDecoratorTest` (unset, partially-set, explicit-value-preserved, init containers, `StatefulSet`, non-Pod no-op). `gradle build` passes.

Generated-by: Claude Code (Opus 4.8)

### DIFF
--- a/spark-operator/src/main/java/org/apache/spark/k8s/operator/decorators/SecurityContextDecorator.java
+++ b/spark-operator/src/main/java/org/apache/spark/k8s/operator/decorators/SecurityContextDecorator.java
@@ -1,0 +1,74 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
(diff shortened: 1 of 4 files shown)


viirya (approved):

dongjoon-hyun:
Thank you always, @viirya !

dongjoon-hyun:
Merged to main

PR #711, opened 2026-06-16
[SPARK-57497] Fix NPE in AppInitStep restart backoff check when stateTransitionHistory is null

jiangzho:

  In AppInitStep.reconcile(), when a previous attempt summary exists, the restart backoff
  check now handles the case where previousAttemptSummary.getStateTransitionHistory() is
  null (non-trim mode). Instead of dereferencing it directly (causing an NPE), the code
  falls back to the main status.getStateTransitionHistory() and selects the state entered
  immediately before the current initializing state — the stopping state that triggered the
  restart — using NavigableMap.lowerEntry().

  A null guard for getRestartConfig() was also added to the same block.

  In non-trim mode (TRIM_ATTEMPT_STATE_TRANSITION_HISTORY=false),
  ApplicationStatus.terminateOrRestart() sets previousAttemptSummary to the old
  currentAttemptSummary, which always carries stateTransitionHistory = null. The previous
  code called .getStateTransitionHistory().get(...) unconditionally, throwing a
  NullPointerException on every restart attempt in non-trim mode. This caused the operator
  to fail to re-initialize the driver after a restart.

  No functional behavior change for users running in trim mode (the default). In non-trim
  mode, restart backoff is now correctly enforced instead of crashing with an NPE.

  Two unit tests were added to AppInitStepTest:

  - nonTrimModeRestartBackoffElapsedProceedsToDriverCreation: verifies that when the backoff has elapsed, the app reaches DriverRequested state without throwing an NPE (non-trim mode, previousAttemptSummary present with null stateTransitionHistory).
  - nonTrimModeRestartBackoffActiveRequeuesWithDelay: verifies that when the backoff has not yet elapsed, the reconcile returns a delayed requeue without throwing an NPE, and the state remains ScheduledToRestart.

  Generated-by: Claude Sonnet 4.6

### DIFF
--- a/spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppInitStep.java
+++ b/spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppInitStep.java
@@ -85,15 +85,14 @@ public final class AppInitStep extends AppReconcileStep {
       }
       RestartConfig restartConfig =
           app.getSpec().getApplicationTolerations().getRestartConfig();
-      if (restartConfig != null) {
-        Instant restartTime =
-            lastTransitionTime.plusMillis(
-                restartConfig.getEffectiveRestartBackoffMillis(
-                    lastState.getCurrentStateSummary()));
-        Instant now = Instant.now();
-        if (restartTime.isAfter(now)) {
-          return ReconcileProgress.completeAndRequeueAfter(Duration.between(now, restartTime));
-        }
+      assert restartConfig != null : "restartConfig must not be null";
+      Instant restartTime =
+          lastTransitionTime.plusMillis(
+              restartConfig.getEffectiveRestartBackoffMillis(
+                  lastState.getCurrentStateSummary()));
+      Instant now = Instant.now();
+      if (restartTime.isAfter(now)) {
+        return ReconcileProgress.completeAndRequeueAfter(Duration.between(now, restartTime));
       }
     }
     try {
--- a/spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppInitStep.java
+++ b/spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppInitStep.java
@@ -27,7 +27,10 @@ import static org.apache.spark.k8s.operator.utils.SparkExceptionUtils.buildGener
 import java.time.Duration;
 import java.time.Instant;
 import java.util.List;
+import java.util.Map;
+import java.util.NavigableMap;
 import java.util.Optional;
+import java.util.SortedMap;
 
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.Pod;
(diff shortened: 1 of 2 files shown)


spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppInitStep.java
```diff
@@ -66,17 +70,30 @@ public ReconcileProgress reconcile(
...
+      final ApplicationState lastState;
+      if (attemptHistory != null && !attemptHistory.isEmpty()) {
+        lastState = attemptHistory.get(attemptHistory.lastKey());
+      } else {
+        // Non-trim mode: previousAttemptSummary carries null stateTransitionHistory.
+        // Fall back to the main history and pick the state entered just before the current
+        // initializing state (which is the stopping state that triggered the restart).
+        NavigableMap<Long, ApplicationState> mainNav =
+            (NavigableMap<Long, ApplicationState>) app.getStatus().getStateTransitionHistory();
+        Map.Entry<Long, ApplicationState> prevEntry = mainNav.lowerEntry(mainNav.lastKey());
```
peter-toth, line 83: Nit: This "entry immediately before the last main-history entry" computation already exists in `AppCleanUpStep.getLastObservedStateBeforeTermination` (`AppCleanUpStep.java:300-309`) — same unchecked `NavigableMap` cast, same `lowerEntry`. The two will drift over time. Consider promoting a single guarded helper onto `ApplicationStatus` and calling it from both sites:

```java
/** State entered immediately before the current (last) one, or the current state if none. */
public ApplicationState getStateBeforeCurrent() {
  NavigableMap<Long, ApplicationState> nav =
      (NavigableMap<Long, ApplicationState>) stateTransitionHistory;
  Map.Entry<Long, ApplicationState> prev = nav.lowerEntry(nav.lastKey());
  return prev != null ? prev.getValue() : nav.lastEntry().getValue();
}
```

Then this branch becomes `lastState = app.getStatus().getStateBeforeCurrent();`. As a bonus it hardens the `AppCleanUpStep` copy, which dereferences `lowerEntry(...).getValue()` with no null guard.

peter-toth (approved):
LGTM, just a minor nit.

spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppInitStep.java
```diff
@@ -66,17 +70,30 @@ public ReconcileProgress reconcile(
...
+        lastState = prevEntry != null ? prevEntry.getValue() : mainNav.lastEntry().getValue();
+      }
+      RestartConfig restartConfig =
+          app.getSpec().getApplicationTolerations().getRestartConfig();
+      if (restartConfig != null) {
```
dongjoon-hyun: May I ask when `restartConfig == null` here? 

If it doesn't happen in the context, can we simply do some assertion statement for this condition instead of `if (restartConfig != null)`

dongjoon-hyun (approved):
+1, LGTM (except only one nit coding style comment.)

dongjoon-hyun (approved):

dongjoon-hyun:
Merged to main

dongjoon-hyun:
Thank you, @jiangzho and Peter.

PR #728, opened 2026-07-01
[SPARK-57845] Exclude `org.roaringbitmap` dependency

dongjoon-hyun:

This PR excludes the transitive `org.roaringbitmap:RoaringBitmap` dependency from the shaded operator jar by adding `exclude group: "org.roaringbitmap"` to the Spark dependency declarations in the `spark-operator` and `spark-submission-worker` modules.

`RoaringBitmap` is pulled in transitively via `org.apache.spark:spark-network-common`, where Spark uses it for shuffle map status tracking (`HighlyCompressedMapStatus`). The operator only builds driver/submission specs and does not run the shuffle/executor runtime, so this library is unused. It belongs to the same category of shuffle/storage runtime libraries that are already excluded (`org.lz4`, `org.rocksdb`, `org.fusesource.leveldbjni`, `org.xerial.snappy`, `com.github.luben`). Removing it keeps the shaded jar lean and reduces the dependency surface.

No.

Pass the CIs.

Generated-by: Claude Opus 4.8

### DIFF
--- a/spark-operator/build.gradle
+++ b/spark-operator/build.gradle
@@ -49,6 +49,7 @@ dependencies {
     exclude group: "org.checkerframework"
     exclude group: "org.fusesource.leveldbjni"
     exclude group: "org.lz4"
+    exclude group: "org.roaringbitmap"
     exclude group: "org.rocksdb"
     exclude group: "org.slf4j"
     exclude group: "org.xerial.snappy"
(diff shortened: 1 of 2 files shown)


viirya (approved):

peter-toth (approved):

dongjoon-hyun:
Thank you, @viirya and @peter-toth .

dongjoon-hyun:
Merged to main

PR #747, opened 2026-07-06
[SPARK-57968] Use virtual threads for `SentinelManager` scheduled executor

dongjoon-hyun:

This PR aims to use virtual threads for the `SentinelManager` scheduled executor by providing a virtual thread factory.

- The sentinel health-check workers spend most of their time on Kubernetes API calls (network I/O), which is an ideal workload for virtual threads.

- In addition, the worker threads now have meaningful names (`sentinel-0`, `sentinel-1`, ...) instead of the anonymous default (`pool-N-thread-M`), which helps thread-dump analysis.

- Note on behavior change: the workers change from non-daemon platform threads to daemon virtual threads (virtual threads are always daemon). This does not affect JVM liveness in practice because the operator process is kept alive by the `com.sun.net.httpserver` dispatcher thread (non-daemon) and the Java Operator SDK threads.

No.

Pass the CIs.

Generated-by: Claude Fable 5


peter-toth (approved):

dongjoon-hyun:
Thank you, @peter-toth !

dongjoon-hyun:
Merged to main

PR #749, opened 2026-07-06
[SPARK-57868] Requeue after driver not found because of the stale cache

TQJADE:
When the operator observes a SparkApplication and the driver pod is absent from the informer (secondary-resource) cache, it previously concluded immediately that the driver had been removed and transitioned the application to Failed (driverUnexpectedRemoved, message "Driver removed…").
  
  This PR makes that determination robust against transient informer-cache staleness:

  - Grace period before verification. If the driver pod is missing from the cache but the application's current state is younger than spark.kubernetes.operator.reconciler.missingDriverGracePeriodSeconds (default 60), the operator defers the verdict and requeues, giving the informer time to catch up (e.g. right after pod creation or a watch reconnect).
  - Authoritative live lookup. Once past the grace period, the operator performs a direct, cache-bypassing lookup against the API server via the new SparkAppContext.getDriverPodFromApi(). Only if the API server also reports no driver pod is the application marked driverUnexpectedRemoved. If the pod is present, the informer is treated as stale and the live pod is used for the current reconcile.
  - Short-lived requeue cadence. Deferrals requeue after spark.kubernetes.operator.reconciler.defaultRequeueIntervalSeconds
  (default 15), distinct from the coarse steady-state reconciler.intervalSeconds (default 120), so the transient condition is re-checked promptly rather than waiting for the next periodic reconcile.
  - API-failure handling. If the live lookup itself throws (API server unreachable), the removal verdict is deferred and the reconcile requeues rather than failing the application; the condition self-heals when the API server recovers.

  Two new dynamically-overridable config options are added in SparkOperatorConf, documented in docs/config_properties.md.

An informer cache can briefly diverge from API-server state. The driver pod appears "missing" from the cache even though it is healthy on the API server. The previous logic treated a cache miss as ground truth and failed the application.

yes

unit test

Yes. Generated-by: Claude Code Opus 4.8 (1M context)


docs/config_properties.md
```diff
@@ -38,9 +38,11 @@
  | spark.kubernetes.operator.periodicGC.intervalSeconds | Long | 1800 | true | Interval (in seconds) between periodic System.gc() invocations. Set to 0 or a negative value to disable. Note that System.gc() is a no-op if JVM is started with -XX:+DisableExplicitGC. | 
  | spark.kubernetes.operator.reconciler.appStatusListenerClassNames | String |  | false | Comma-separated names of SparkAppStatusListener class implementations | 
  | spark.kubernetes.operator.reconciler.clusterStatusListenerClassNames | String |  | false | Comma-separated names of SparkClusterStatusListener class implementations | 
+ | spark.kubernetes.operator.reconciler.defaultRequeueIntervalSeconds | Integer | 15 | true | Default requeue interval (in seconds) used for short-lived requeues while waiting for a transient condition to clear, e.g., for the pod informer cache to catch up during the missing-driver grace period. Distinct from spark.kubernetes.operator.reconciler.intervalSeconds, which controls the steady-state periodic reconcile cadence. |
```
dongjoon-hyun: `defaultRequeueIntervalSeconds` is very misleading because it's not the real default. It's only used for missing driver cases.

The real default is `reconciler.intervalSeconds` in `completeAndDefaultRequeue` method.

https://github.com/apache/spark-kubernetes-operator/blob/1127bf9fee08b55251401b86c847e6646b520c98/spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/ReconcileProgress.java#L65-L68

dongjoon-hyun (changes requested):
Please revise the configuration name.

spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppReconcileStep.java
```diff
@@ -65,30 +67,69 @@ protected ReconcileProgress observeDriver(
...
+        } else {
+          ApplicationStatus updatedStatus =
+              currentStatus.appendNewState(driverUnexpectedRemoved());
+          return attemptStatusUpdate(
+              context, statusRecorder, updatedStatus, completeAndImmediateRequeue());
         }
-        return attemptStatusUpdate(
-            context, statusRecorder, currentStatus, completeAndImmediateRequeue());
+      } catch (RuntimeException e) {
+          log.error(
```
dongjoon-hyun: Indentation?

spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppReconcileStep.java
```diff
@@ -65,30 +67,69 @@ protected ReconcileProgress observeDriver(
...
+          driverPodOptional = liveDriver;
+        } else {
+          ApplicationStatus updatedStatus =
+              currentStatus.appendNewState(driverUnexpectedRemoved());
+          return attemptStatusUpdate(
+              context, statusRecorder, updatedStatus, completeAndImmediateRequeue());
         }
-        return attemptStatusUpdate(
-            context, statusRecorder, currentStatus, completeAndImmediateRequeue());
+      } catch (RuntimeException e) {
```
dongjoon-hyun: Please use the narrowest one. Probably, `KubernetesClientException`.

docs/config_properties.md
TQJADE: Changed to `spark.kubernetes.operator.reconciler.missingDriverRequeueIntervalSeconds`

spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppReconcileStep.java
TQJADE: updated.

spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppReconcileStep.java
TQJADE: Replaced with KubernetesClientException.

dongjoon-hyun:
Thank you, @TQJADE .

TQJADE:
> Please revise the configuration name.

https://github.com/apache/spark-kubernetes-operator/pull/749#discussion_r3547149716

docs/config_properties.md
```diff
@@ -35,12 +35,14 @@
...
+ | spark.kubernetes.operator.reconciler.missingDriverGracePeriodSeconds | Integer | 60 | true | Grace period (in seconds) the operator waits for the pod informer cache to catch up after the driver pod is observed missing, before doing a live API verification against the Kubernetes apiserver. This protects healthy applications from being failed when the informer cache is briefly stale (e.g., after a watch reconnect). Set to 0 to verify on every cache miss. | 
+ | spark.kubernetes.operator.reconciler.missingDriverRequeueIntervalSeconds | Integer | 15 | true | Requeue interval (in seconds) used to re-check a driver pod that is missing from the informer cache, while waiting for the cache to catch up during the missing-driver grace period. Specific to the missing-driver flow; this is not the operator's default requeue cadence. The steady-state periodic reconcile interval is spark.kubernetes.operator.reconciler.intervalSeconds. |
```
dongjoon-hyun: What happens for `0`? It would be great to describe like `spark.kubernetes.operator.reconciler.missingDriverGracePeriodSeconds`.

Also, for the negative value, could you double-check the behavior?

spark-operator/src/main/java/org/apache/spark/k8s/operator/config/SparkOperatorConf.java
```diff
@@ -220,6 +199,47 @@ public final class SparkOperatorConf {
...
+              "Requeue interval (in seconds) used to re-check a driver pod that is missing from "
+                  + "the informer cache, while waiting for the cache to catch up during the "
+                  + "missing-driver grace period. Specific to the missing-driver flow; this is "
+                  + "not the operator's default requeue cadence. The steady-state periodic "
+                  + "reconcile interval is spark.kubernetes.operator.reconciler.intervalSeconds.")
+          .typeParameterClass(Integer.class)
+          .defaultValue(15)
+          .build();
+
+
```
dongjoon-hyun: Redundant empty line.

spark-operator/src/main/java/org/apache/spark/k8s/operator/context/SparkAppContext.java
```diff
@@ -70,6 +70,23 @@ public Optional<Pod> getDriverPod() {
...
+    return josdkContext
+        .getClient()
+        .pods()
+        .inNamespace(sparkApplication.getMetadata().getNamespace())
+        .withLabels(driverLabels(sparkApplication))
+        .list()
+        .getItems()
+        .stream()
+        .findAny();
+  }
```
dongjoon-hyun, line 98: This could be a regression due to the newly added excessive API calls.

In states like `DriverRequested`/`DriverStarted`, two or three `AppResourceObserveStep`s run back-to-back in a single reconcile loop, so while the informer cache is stale each step performs its own live API-server lookup (and emits its own warn log). Since `SparkAppContext` is shared by all steps within one reconcile loop, memoizing the lookup result — following the same lazy-initialization pattern as `secondaryResourceSpec` — reduces this to a single API call per loop. (If the lookup throws `KubernetesClientException`, nothing is cached, so the next step naturally retries.)

```suggestion
  /** Memoized live driver pod lookup - initialized upon the first attempt in this reconcile. */
  private Optional<Pod> driverPodFromApi;

  /**
   * Live lookup of the driver pod against the API server, bypassing the informer cache. The
   * result is memoized so that multiple reconcile steps within the same reconcile loop share a
   * single API call.
   *
   * @return An Optional containing the driver Pod as known to the API server.
   */
  public Optional<Pod> getDriverPodFromApi() {
    synchronized (this) {
      if (driverPodFromApi == null) {
        driverPodFromApi =
            josdkContext
                .getClient()
                .pods()
                .inNamespace(sparkApplication.getMetadata().getNamespace())
                .withLabels(driverLabels(sparkApplication))
                .list()
                .getItems()
                .stream()
                .findAny();
      }
      return driverPodFromApi;
    }
  }
```

dongjoon-hyun:
I finished another round of review, @TQJADE .

docs/config_properties.md
TQJADE: updated.

spark-operator/src/main/java/org/apache/spark/k8s/operator/config/SparkOperatorConf.java
TQJADE: removed

spark-operator/src/main/java/org/apache/spark/k8s/operator/context/SparkAppContext.java
TQJADE: add lazy initialization.

dongjoon-hyun:
Thank you for updating, @TQJADE .

dongjoon-hyun (approved):
+1, LGTM.

dongjoon-hyun:
cc @jiangzho

dongjoon-hyun:
Merged to main.

PR #764, opened 2026-07-23
[SPARK-58284] Remove duplicated NetworkPolicy creation in favor of `NetworkPolicyFeatureStep` of Spark 4.2

dongjoon-hyun:

This PR aims to remove the operator-side `NetworkPolicy` creation from `SparkAppResourceSpec` in favor of Apache Spark 4.2.0's built-in `NetworkPolicyFeatureStep`.

- Remove `buildNetworkPolicy` and its invocation from `SparkAppResourceSpec`.
- Remove the corresponding `testNetworkPolicy` test case and adjust the pre-resource count assertion.
- Keep the `networkpolicies` RBAC rule in the Helm chart because the operator still needs it to create the `NetworkPolicy` generated by Spark's `NetworkPolicyFeatureStep`.

The operator has created a `NetworkPolicy` per `SparkApplication` since SPARK-55085. After SPARK-58194 upgraded the Spark dependency to 4.2.0, Spark's `KubernetesDriverBuilder` also generates a functionally identical `NetworkPolicy` via `NetworkPolicyFeatureStep` (SPARK-55653). As a result, every `SparkApplication` ends up with two duplicated policies.

- https://github.com/apache/spark/pull/54442

```
$ kubectl get networkpolicy
NAME                  POD-SELECTOR                                  AGE
pi-0-network-policy   spark-app-selector=pi-0,spark-role=executor   41s
pi-0-policy           spark-app-selector=pi-0,spark-role=executor   40s
```

Both policies have the same pod selector and ingress rule (allowing traffic only from the pods of the same Spark application). Delegating to Spark's `NetworkPolicyFeatureStep` removes the duplication and avoids maintaining the same logic in two places.

Yes, only one `NetworkPolicy` (`<appId>-policy`, created by Spark) is deployed per `SparkApplication` instead of two. The security behavior is unchanged because the two policies were functionally identical.

Pass the CIs.

Generated-by: Claude Fable 5

### DIFF
--- a/spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppResourceSpec.java
+++ b/spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkAppResourceSpec.java
@@ -27,8 +27,6 @@ import io.fabric8.kubernetes.api.model.ContainerBuilder;
 import io.fabric8.kubernetes.api.model.HasMetadata;
 import io.fabric8.kubernetes.api.model.Pod;
 import io.fabric8.kubernetes.api.model.PodBuilder;
-import io.fabric8.kubernetes.api.model.networking.v1.NetworkPolicy;
-import io.fabric8.kubernetes.api.model.networking.v1.NetworkPolicyBuilder;
 import lombok.Getter;
 
 import org.apache.spark.deploy.k8s.Config;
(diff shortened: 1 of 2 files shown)


viirya (approved):

peter-toth (approved):

dongjoon-hyun:
Thank you, @viirya and @peter-toth !

dongjoon-hyun:
Merged to main

PR #773, opened 2026-07-23
[SPARK-58306] Use `ubuntu-26.04(-arm)?` in GitHub Action jobs

dongjoon-hyun:

This PR aims to use `ubuntu-26.04` and `ubuntu-26.04-arm` runner images in all GitHub Action jobs.

- `ubuntu-slim` → `ubuntu-26.04` (This is better because `ubuntu-slim` is flaky in these days due to 15-min timeout)
- `ubuntu-latest` → `ubuntu-26.04`
- `ubuntu-24.04-arm` → `ubuntu-26.04-arm`

For Apache Spark K8s Operator v1.1.0, this PR aims the following:

1. **Consistent environment across the build matrix** — Previously, the x86 job used `ubuntu-slim` (24.04-based) while the arm job used `ubuntu-24.04-arm`, so the two architectures ran on different image types. Now both architectures are tested on the same Ubuntu 26.04 environment, eliminating variables other than the architecture itself.

2. **Longer support window** — The 26.04 runner image will be supported longer than the 24.04-based images, buying time until the next LTS migration.

3. **Early validation on the latest kernel and system libraries** — The operator is built and tested on top of the newest glibc, kernel, and Docker versions, so issues on modern environments are caught in CI first.

No. This is an infra-only change.

Pass the CIs.

Generated-by: Claude Fable 5

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -6,7 +6,7 @@ on:
 jobs:
   license-check:
     name: "License Check"
-    runs-on: ubuntu-slim
+    runs-on: ubuntu-26.04
     timeout-minutes: 20
     steps:
       - name: Checkout repository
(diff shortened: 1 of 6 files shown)


viirya (approved):

dongjoon-hyun:
Thank you so much always, @viirya !

dongjoon-hyun:
Merged to main

PR #778, opened 2026-07-29
[SPARK-58426] Render each JVM option as its own command element

snicholasbarton:

Each JVM option in `operatorDeployment.operatorPod.operatorContainer.jvmArgs` now renders as its own element of the operator container's `command`, instead of the whole string being passed through a single `"$(OPERATOR_JAVA_OPTS)"` element. The `with` guard drops empty strings, so leading, trailing, and repeated spaces produce no empty arguments.

Also adds a helm-unittest suite covering the rendered `command`, runs it from the existing lint job, and anchors `/tests/` in `.helmignore` so the suite is not packaged into the released chart.

Kubernetes `$(VAR)` expansion substitutes into a single argv element and never splits on whitespace, so the entire `jvmArgs` string reaches the JVM as one argument. It begins with `-D`, so the JVM reads it as one system property and ignores every later option. Nothing errors or warns.

None of the shipped defaults (`CrashOnOutOfMemoryError`, `ErrorFile`, `UseParallelGC`, `InitialRAMPercentage`, `MaxRAMPercentage`, `AlwaysPreTouch`, and `UseCompactObjectHeaders`) have any effect. E.g. the operator runs at the JVM default 25% `MaxRAMPercentage` instead of 80%.

Yes, insofar as we fixed a bug: `jvmArgs` values now take effect as expected.

Pass the CIs.

Generated-by: Claude Code (claude-opus-5)


dongjoon-hyun (changes requested):
Thank you for the fix, @snicholasbarton . However, Apache Spark community doesn't want to depend on 3rd-party external repository, `helm-unittest`.

snicholasbarton:
Thanks for taking a look @dongjoon-hyun , that's fair feedback. I've ported the tests I added to use native Helm by pattern-matching on the structure of the existing tests. Let me know how you feel about this, and if you want any other changes.

dongjoon-hyun (approved):
+1, LGTM. Thank you, @snicholasbarton .

dongjoon-hyun:
Merged to main for Apache Spark K8s Operator v1.1.0.

dongjoon-hyun:
Congratulations for your first commit, @snicholasbarton .

I added you (JIRA ID: `snicholasbarton`) to the Apache Spark contributor group and assigned SPARK-58426 to you.

Welcome to the Apache Spark community!

PR #783, opened 2026-08-04
[SPARK-58565] Run every assertion in the Helm RBAC test hooks

snicholasbarton:
Move the RBAC Helm test assertions into a single block scalar, one `kubectl auth can-i` per line under `set -ex`. `can-i` already exits non-zero when it answers "no", and -x echoes each assertion into the pod log beside its answer. 

Expand the test assertions to more comprehensively cover cases not already included:

- Cover every grant; only core resources and spark.apache.org were checked before. 
  - Leases are asserted only for a replicated operator, matching the replicas > 1 condition on that grant.
  - Assert the Gateway API grant only where the CRDs are installed. `kubectl auth can-i` answers "no" for a resource type the cluster does not serve, even when the rules grant it, so an unguarded assertion would fail on every cluster without the Gateway API.
  - Add negative assertions
- Stop deleting the pod when a hook fails, so its log survives for inspection. A passing pod is still cleaned up.

If a future change breaks the permissions model, the tests will continue to pass. We should catch breaking changes in tests.

`command: ['bash', '-c']` with a multi-element `args` array passes only the first element to bash as the script; the rest become positional parameters. Six of the seven operator assertions and three of the four workload assertions therefore never ran, and `helm test` passed regardless.

No

Repro listed in the ticket + it should pass CI.

Generated by: Claude Code (Opus 5)

### DIFF
--- a/build-tools/helm/spark-kubernetes-operator/templates/tests/test-rbac.yaml
+++ b/build-tools/helm/spark-kubernetes-operator/templates/tests/test-rbac.yaml
@@ -79,10 +79,11 @@ spec:
           fi
 
           # The operator must not be cluster-admin. These also prove the suite is
-          # able to observe a denial at all.
-          ! kubectl auth can-i create clusterroles.rbac.authorization.k8s.io
-          ! kubectl auth can-i delete nodes
-          ! kubectl auth can-i '*' '*' --all-namespaces
+          # able to observe a denial at all. `set -e` ignores a failing `!`
+          # pipeline, so each denial must be asserted with an explicit exit.
+          if kubectl auth can-i create clusterroles.rbac.authorization.k8s.io; then exit 1; fi
+          if kubectl auth can-i delete nodes; then exit 1; fi
+          if kubectl auth can-i '*' '*' --all-namespaces; then exit 1; fi
   serviceAccountName: {{ .Values.operatorRbac.serviceAccount.name }}
   restartPolicy: Never
 ---
@@ -116,8 +117,9 @@ spec:
           kubectl auth can-i watch  statefulsets.apps --namespace={{ .Release.Namespace }}
           kubectl auth can-i create statefulsets.apps --namespace={{ .Release.Namespace }}
 
-          # Spark workloads must not reach beyond their own namespace.
-          ! kubectl auth can-i create clusterroles.rbac.authorization.k8s.io
-          ! kubectl auth can-i '*' '*' --all-namespaces
+          # Spark workloads must not reach beyond their own namespace. `set -e`
+          # ignores a failing `!` pipeline, so assert each denial explicitly.
+          if kubectl auth can-i create clusterroles.rbac.authorization.k8s.io; then exit 1; fi
+          if kubectl auth can-i '*' '*' --all-namespaces; then exit 1; fi
   serviceAccountName: {{ .Values.workloadResources.serviceAccount.name }}
   restartPolicy: Never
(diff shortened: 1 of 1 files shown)


snicholasbarton:
TBH I didn't think this warranted a JIRA ticket but made one anyway to be on the safe side since it seems like that's the general practice here. Are tickets a hard requirement for this project, even for these minor changes to e.g. test harnesses?

build-tools/helm/spark-kubernetes-operator/templates/tests/test-rbac.yaml
```diff
@@ -21,21 +21,68 @@ metadata:
...
+            kubectl auth can-i create httproutes.gateway.networking.k8s.io --all-namespaces
+            kubectl auth can-i watch  grpcroutes.gateway.networking.k8s.io --all-namespaces
+            kubectl auth can-i create grpcroutes.gateway.networking.k8s.io --all-namespaces
+          fi
+
+          # The operator must not be cluster-admin. These also prove the suite is
+          # able to observe a denial at all.
+          ! kubectl auth can-i create clusterroles.rbac.authorization.k8s.io
+          ! kubectl auth can-i delete nodes
+          ! kubectl auth can-i '*' '*' --all-namespaces
```
dongjoon-hyun: `set -e` does not fire on a pipeline preceded by `!` (POSIX/bash: "the shell does not exit if the failed command is part of a pipeline preceded by the `!` reserved word"). So if the SA *can* create clusterroles or delete nodes, these two lines return 1 but the script keeps going — only the last line is enforced, and only because its status happens to be the script's exit status. That's the same "assertion that never fails" class of bug this PR is fixing.

An `if ...; then exit 1; fi` makes each denial mandatory. (Note `can-i X && exit 1` doesn't work either: on a denial the list itself exits 1, which would fail the pod when it's the last command.)

```suggestion
          # able to observe a denial at all. `set -e` ignores a failing `!`
          # pipeline, so each denial must be asserted with an explicit exit.
          if kubectl auth can-i create clusterroles.rbac.authorization.k8s.io; then exit 1; fi
          if kubectl auth can-i delete nodes; then exit 1; fi
          if kubectl auth can-i '*' '*' --all-namespaces; then exit 1; fi
```

build-tools/helm/spark-kubernetes-operator/templates/tests/test-rbac.yaml
```diff
@@ -48,17 +95,29 @@ metadata:
...
+          kubectl auth can-i watch  configmaps --namespace={{ .Release.Namespace }}
+          kubectl auth can-i create configmaps --namespace={{ .Release.Namespace }}
+          kubectl auth can-i watch  persistentvolumeclaims --namespace={{ .Release.Namespace }}
+          kubectl auth can-i create persistentvolumeclaims --namespace={{ .Release.Namespace }}
+          kubectl auth can-i watch  statefulsets.apps --namespace={{ .Release.Namespace }}
+          kubectl auth can-i create statefulsets.apps --namespace={{ .Release.Namespace }}
+
+          # Spark workloads must not reach beyond their own namespace.
+          ! kubectl auth can-i create clusterroles.rbac.authorization.k8s.io
+          ! kubectl auth can-i '*' '*' --all-namespaces
```
dongjoon-hyun: Same issue as above: under `set -e` a `!`-prefixed command that fails does not stop the script, so the `clusterroles` assertion here is never enforced.

```suggestion
          # Spark workloads must not reach beyond their own namespace. `set -e`
          # ignores a failing `!` pipeline, so assert each denial explicitly.
          if kubectl auth can-i create clusterroles.rbac.authorization.k8s.io; then exit 1; fi
          if kubectl auth can-i '*' '*' --all-namespaces; then exit 1; fi
```

dongjoon-hyun (changes requested):
Thank you, @snicholasbarton . Could you revise the PR?

snicholasbarton:
Thanks for the review @dongjoon-hyun . TIL! Have applied both suggestions.

dongjoon-hyun (approved):
+1, LGTM. Thank you!

dongjoon-hyun:
Merged to main.

PR #794, opened 2026-08-10
[SPARK-58691] Add configurable worker metrics NetworkPolicy ingress

yalindogusahin:

Add an optional `workerSpec.networkPolicy` block to `SparkCluster` with a metrics port and a list of `NetworkPolicyPeer` sources. The generated worker policy admits those peers on the configured TCP port, alongside the existing cluster and driver rules. An omitted block adds no metrics ingress rule.

The CRD requires both fields when the block is present, restricts the port to 1-65535, and requires at least one peer. The operator does not check whether the configured port serves the worker UI; users should choose a dedicated exporter port.

Include a JMX Exporter example using `SPARK_DAEMON_JAVA_OPTS` and Spark's `JmxSink`, document the custom-image prerequisite, and assert the generated policy in the existing state-transition e2e suite.

The generated worker policy admits cluster and driver pods by default. Prometheus scrapers without those labels cannot reach a worker metrics endpoint. Configurable port and peer rules allow administrators to admit their scrapers to a dedicated exporter endpoint.

Yes. Users can configure metrics ingress under `workerSpec.networkPolicy`:

```yaml
spec:
  workerSpec:
    networkPolicy:
      metricsPort: 9404
      metricsIngress:
        - namespaceSelector:
            matchLabels:
              kubernetes.io/metadata.name: monitoring
```

The API server rejects incomplete blocks, empty peer lists, and out-of-range ports. Users must configure the exporter separately; this block configures network access only. NetworkPolicy enforcement requires a compatible CNI plugin.

```text
./gradlew spotlessApply :spark-operator-api:generateCrds
./gradlew build javadoc
helm lint --strict build-tools/helm/spark-kubernetes-operator
```

The unit tests cover default ingress, explicit metrics peers, missing ports, empty peer lists, and policy apply failures. The Gradle build checks that the generated CRD matches the Helm chart.

I also scraped a local Spark 4.1.2 worker with JMX Exporter 1.6.0, using the example ConfigMap and JVM settings with the exporter jar mounted. The endpoint returned `spark_worker_coresFree`, `spark_worker_coresUsed`, `spark_worker_executors`, `spark_worker_memFree_MB`, and `spark_worker_memUsed_MB`. This did not exercise Kubernetes NetworkPolicy or Spark 4.2.0.

The policy assertion is part of the state-transition e2e suite, which CI already runs. E2e execution was not performed locally because Chainsaw is not installed.

Generated-by: Codex (GPT-6)


dongjoon-hyun (changes requested):
Thank you for making a PR, @yalindogusahin . 

However, I'm -1 for this AS-IS PR because 8081 is used for the whole worker UI.

yalindogusahin:
Thank you for reviewing and that is correct my bad at all I haven't checked the metric names.

However, if we decide to use jmx exporter to expose spark related metrics we should enable this network policy. For the statsd thats not the case I believe.

If we decide to go with jmx exporter we should bake jmx exporter to the image itself and + mapping config.
Looking forward to hear you opinion about this.

yalindogusahin:
Pushed a redesign that addresses this directly: added an optional `workerSpec.metricsPort` field. When set, the generated `NetworkPolicy` opens ingress on *that specific port only* — not port 8081/the web UI. The idea is to pair this with a JMX-to-Prometheus exporter agent (`jmx_prometheus_javaagent`) attached to the worker JVM as a `-javaagent`, which serves metrics on its own dedicated port, fully decoupled from the web UI. Default behavior (no `metricsPort` set) is unchanged from before this PR — the web UI port is never touched. Added docs (`docs/operations.md`, new "Exposing SparkCluster Worker Metrics" section) and a runnable example (`examples/cluster-with-jmx-exporter.yaml`) showing the full setup. Let me know if this addresses your concern or if you'd prefer a different shape for the config.

dongjoon-hyun (commented):
Thank you for updating the PR. Could you change the direction like the following? Apache Spark Operator Pod itself also has the similar feature in a better way. If we want to have this feature, we need to use ingress feature instead of opening the port to the world. WDYT?

https://github.com/apache/spark-kubernetes-operator/blob/e03310d2cb28deb2758d37d25a810eaf9cb49cd9/docs/operations.md?plain=1#L158

yalindogusahin:
Thank you for the pointer. It makes sense to restrict the metrics port to explicitly selected peers rather than opening it up, so I have switched to that direction.

The generated worker `NetworkPolicy ` now carries a from: peer list supplied by the user under a new `workerSpec.metricsIngress` field, deliberately named and typed to match `operatorDeployment.networkPolicy.metricsIngress` in the chart, so the two features read the same way:

```yaml
spec:
  workerSpec:
    metricsPort: 9404
    metricsIngress:
      - namespaceSelector:
          matchLabels:
            kubernetes.io/metadata.name: "monitoring"
```

Both fields are required before any rule is generated: the port alone would admit every source, and the peers alone would grant them every worker port. When either is missing, worker ingress is byte-for-byte what it is on main.

One naming question: `metricsIngress` echoes the chart, but ingress may read as the Kubernetes Ingress resource rather than a NetworkPolicy peer list. I am happy to rename it if you consider that clearer.

dongjoon-hyun:
Thank you for the update, @yalindogusahin. The new direction (dedicated port + explicit `NetworkPolicyPeer` list, nothing generated unless both are set) looks right to me, and the generated policy is exactly what I would expect:

```yaml
ingress:
- from:
  - podSelector:
      matchLabels:
        spark.operator/spark-cluster-name: cluster-with-jmx-exporter
  - podSelector:
      matchLabels:
        spark-role: driver
- from:
  - namespaceSelector:
      matchLabels:
        kubernetes.io/metadata.name: monitoring
  ports:
  - port: 9404
    protocol: TCP
```

I built the PR locally and `:spark-operator-api:build`, `:spark-submission-worker:build`, `javadoc`, and `assertGeneratedCRDMatchesHelmChart` all pass, so the CRD and the Helm chart are in sync.

However, there is one blocker before this can go in.

## The documented setup does not actually work

`docs/operations.md` (step 3) and `examples/cluster-with-jmx-exporter.yaml` both attach the agent through `SPARK_WORKER_OPTS`, but the operator appends its own `SPARK_WORKER_OPTS` to the worker container *after* merging the user's pod template — see `SparkClusterResourceSpec#buildWorkerStatefulSet`:

```java
.editMatchingContainer(p -> "worker".equals(p.getName()))
.withImage(image)
.addNewEnv().withName("SPARK_LOG_DIR")...endEnv()
.addNewEnv().withName("SPARK_WORKER_OPTS").withValue(options).endEnv()
```

Feeding `examples/cluster-with-jmx-exporter.yaml` through `SparkClusterResourceSpec` and dumping the resulting container gives:

```
container=worker
  env SPARK_WORKER_OPTS = -javaagent:/opt/jmx_exporter/jmx_prometheus_javaagent.jar=9404:/etc/metrics/jmx-exporter-config.yaml
  env SPARK_LOG_DIR     = /opt/spark/work/logs
  env SPARK_WORKER_OPTS =
```

kubelet's `makeEnvironmentVariables` collapses duplicate names into a map, so the operator's (here empty) value wins and the user's `-javaagent` is silently dropped. The port would be opened, but nothing would be listening on it.

Please use `SPARK_DAEMON_JAVA_OPTS` instead. Spark's `SparkClassCommandBuilder` adds both `SPARK_DAEMON_JAVA_OPTS` and `SPARK_WORKER_OPTS` for `org.apache.spark.deploy.worker.Worker`, and the operator never sets the former, so there is no collision.

## Example / docs

- The docs call the example "complete, runnable", but the example pulls the stock `apache/spark:{{SPARK_VERSION}}-scala` image, which does not contain `/opt/jmx_exporter/jmx_prometheus_javaagent.jar`. Once the env-var issue above is fixed, a `-javaagent` pointing at a missing jar aborts JVM startup. Please drop the "runnable" wording, or note the custom-image prerequisite next to the image field itself.
- `spark.ui.reverseProxy: "true"` in the example is unrelated to this feature.
- The docs pin `jmx_prometheus_javaagent` 1.0.1; a version-less link would age better.

## API shape

Since these land in the `v1` CRD and are effectively permanent, could we group them?

```yaml
workerSpec:
  networkPolicy:
    metricsPort: 9404
    metricsIngress:
      - namespaceSelector: ...
```

This mirrors the chart's `operatorDeployment.networkPolicy.metricsIngress` nesting more closely, keeps `WorkerSpec` from growing two flat fields that are only meaningful together, and makes it obvious that `metricsPort` configures the policy rather than any metrics endpoint (which is what your own naming question was getting at).

Related: the Javadoc says the web UI port "is deliberately not usable as the metrics port", but nothing rejects `metricsPort: 8081` — it would open the UI to the listed peers. Either validate it or soften the wording.

## Minor

- The branch is 14 commits behind `main`; please rebase.
- No `tests/e2e` coverage for the new fields. Optional, but worth considering for a user-facing CRD field whose effect is only observable at runtime.

yalindogusahin:
Thanks for the detailed review. I went through each point and pushed the fixes.

**Env-var collision**
The docs and example now use `SPARK_DAEMON_JAVA_OPTS` instead of `SPARK_WORKER_OPTS`. The operator appends its own (sometimes empty) `SPARK_WORKER_OPTS` to the worker container after merging the pod template, which silently overwrote the user value. `SPARK_DAEMON_JAVA_OPTS` is read by Spark`s `SparkClassCommandBuilder` for the Worker and the operator never sets it, so nothing collides.

**API shape**
Both fields now sit under `workerSpec.networkPolicy.metricsPort` and `workerSpec.networkPolicy.metricsIngress`, matching `operatorDeployment.networkPolicy.metricsIngress`. It also clears up the naming confusion; `networkPolicy` now reads as a policy block rather than two loose metrics fields.

`metricsPort: 8081` (the web UI port) is now rejected with an `IllegalArgumentException` and a unit test, since the UI and its metrics endpoint share the same embedded HTTP server.

**Docs / example**
Dropped the "complete, runnable" wording and put a note about the custom-image requirement next to the image field. Removed the unrelated `spark.ui.reverseProxy`.

I kept the `jmx_prometheus_javaagent` version in the Dockerfile `ADD` URL because that needs a concrete artifact. Happy to remove it if you would rather not pin it.

**Housekeeping**
Rebased onto the latest main, and added a `tests/e2e/worker-metrics-network-policy/` Chainsaw suite that checks the generated worker NetworkPolicy.

Local build is green: `:spark-operator-api:build`, `:spark-submission-worker:build`, `javadoc`, and `assertGeneratedCRDMatchesHelmChart` all pass, and the CRD is in sync with the Helm chart.

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkClusterResourceSpec.java
```diff
@@ -467,35 +469,67 @@ private static Optional<PodDisruptionBudget> buildPodDisruptionBudget(
...
+            .addToMatchLabels(LABEL_SPARK_ROLE_NAME, LABEL_SPARK_ROLE_DRIVER_VALUE)
+            .endPodSelector()
+            .endFrom()
+            .endIngress();
+    WorkerNetworkPolicySpec networkPolicy = workerSpec.getNetworkPolicy();
+    if (networkPolicy != null
+        && networkPolicy.getMetricsPort() != null
+        && networkPolicy.getMetricsIngress() != null
+        && !networkPolicy.getMetricsIngress().isEmpty()) {
+      if (networkPolicy.getMetricsPort() == 8081) {
```
dongjoon-hyun, line 519: This compares against a literal, not against the worker UI port. Line 77 above forwards every `sparkConf` entry as `-D` into `SPARK_WORKER_OPTS`, and Spark's `WorkerArguments` reads `spark.worker.ui.port` (and `SPARK_WORKER_WEBUI_PORT`), so `spark.worker.ui.port: "9404"` + `metricsPort: 9404` passes here and opens the whole UI to the peers. I would drop the throw and soften the docs/Javadoc instead. If a check is kept, derive it from `conf` and share one constant with the `8081` at lines 190/350.

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkClusterResourceSpec.java
```diff
@@ -467,35 +469,67 @@ private static Optional<PodDisruptionBudget> buildPodDisruptionBudget(
...
+        throw new IllegalArgumentException(
```
dongjoon-hyun, line 520: Throwing from the resource-spec constructor surfaces as a generic `SchedulingFailure` (stack trace in the status message) in `ClusterInitStep`, and from the next reconcile on as `Failed` / `Cannot process cluster status` via `ClusterUnknownStateStep`, which immediate-requeues forever and appends a `Failed` entry to `stateTransitionHistory` each time. Spec validation belongs in `ClusterValidateStep` (currently a no-op placeholder; `AppValidateStep.validateGatewayRoutes` shows the pattern) or in the CRD schema, so this constructor keeps no exception path.

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkClusterResourceSpec.java
```diff
@@ -467,35 +469,67 @@ private static Optional<PodDisruptionBudget> buildPodDisruptionBudget(
...
-        .withNewPodSelector()
-        .addToMatchLabels(LABEL_SPARK_ROLE_NAME, LABEL_SPARK_ROLE_DRIVER_VALUE)
-        .endPodSelector()
-        .endFrom()
-        .endIngress()
-        .endSpec()
-        .build();
+  private NetworkPolicy buildWorkerNetworkPolicy(
+      String clusterName, String namespace, WorkerSpec workerSpec) {
+    var builder =
```
dongjoon-hyun, line 490: Nit: the `builder = builder.addNewIngress()...endIngress();` reassignment below is a no-op. fabric8's `endIngress()` calls `setToIngress` on the enclosing fluent and returns that same instance, so a plain statement works. An early `return builder.endSpec().build();` when the block is absent would also put the rule-add at one nesting level.

spark-operator-api/src/main/java/org/apache/spark/k8s/operator/spec/WorkerNetworkPolicySpec.java
```diff
@@ -0,0 +1,56 @@
...
+@AllArgsConstructor
+@Builder
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public class WorkerNetworkPolicySpec {
+  /**
+   * Port of a metrics endpoint (e.g. a JMX-to-Prometheus exporter agent) running in the worker
+   * container, outside Spark's own web UI. The caller is responsible for making the container
+   * actually listen on this port. Takes effect only together with {@link #metricsIngress}.
+   */
+  protected Integer metricsPort;
```
dongjoon-hyun, line 47: Setting only one of `metricsPort` / `metricsIngress` is silently ignored: no rule, no status, no event. Since this block is optional as a whole, `@Required` on both fields (already used on `RuntimeVersions.sparkVersion`; nested `required:` lists exist under the optional `workerSpec.instanceConfig`) plus `@Min(1) @Max(65535)` on `metricsPort` lets the API server reject a partial or out-of-range block at apply time while leaving an omitted block valid. The null-guard, the "both required" docs paragraph, and `testWorkerNetworkPolicyWith*ButNo*` can then go.

spark-operator-api/src/main/java/org/apache/spark/k8s/operator/spec/WorkerNetworkPolicySpec.java
```diff
@@ -0,0 +1,56 @@
...
+import io.fabric8.kubernetes.api.model.networking.v1.NetworkPolicyPeer;
+import lombok.AllArgsConstructor;
+import lombok.Builder;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+/**
+ * Network policy for the Spark workers.
+ *
+ * @since 0.1.0
```
dongjoon-hyun, line 34: This class is new; `@since 1.1.0`.

examples/cluster-with-jmx-exporter.yaml
```diff
@@ -0,0 +1,77 @@
...
+    rules:
+    - pattern: 'java.lang<type=Memory><HeapMemoryUsage>(\w+)'
+      name: jvm_memory_heap_$1
+      type: GAUGE
+    - pattern: 'java.lang<type=GarbageCollector, name=(\w+)><>CollectionCount'
+      name: jvm_gc_collection_count
+      labels:
+        gc: "$1"
+      type: COUNTER
+    - pattern: 'metrics<name=(\S+)><>Value'
```
dongjoon-hyun, line 30: This rule matches nothing as configured. Spark's `MetricsConfig` registers only the servlet sink by default, so the `metrics` JMX domain stays empty unless `spark.metrics.conf.*.sink.jmx.class: org.apache.spark.metrics.sink.JmxSink` is added to `sparkConf` (it is forwarded to the worker as `-D`). Without it the exporter serves `jvm_*` only and no `spark_worker_*` series.

docs/operations.md
```diff
@@ -174,6 +174,123 @@ Note that this requires a CNI plugin that enforces NetworkPolicy; on clusters wi
...
+   ```yaml
+   apiVersion: v1
+   kind: ConfigMap
+   metadata:
+     name: jmx-exporter-config
+   data:
+     jmx-exporter-config.yaml: |
+       lowercaseOutputName: true
+       rules:
+         - pattern: ".*"
```
dongjoon-hyun, line 212: This catch-all config differs from the targeted rules in `examples/cluster-with-jmx-exporter.yaml`, and it exports every MBean attribute on each scrape. I would keep the full CR in the example only and show just the `networkPolicy` fragment here.

spark-submission-worker/src/test/java/org/apache/spark/k8s/operator/SparkClusterResourceSpecTest.java
```diff
@@ -337,6 +344,62 @@ void testWorkerNetworkPolicy() {
...
   }
 
+  @Test
+  void testWorkerNetworkPolicyWithMetricsIngress() {
+    var scraper = new NetworkPolicyPeerBuilder()
+        .withNewNamespaceSelector()
+        .addToMatchLabels("kubernetes.io/metadata.name", "monitoring")
+        .endNamespaceSelector()
+        .build();
+    var networkPolicySpec = mock(WorkerNetworkPolicySpec.class);
```
dongjoon-hyun, line 354: `WorkerNetworkPolicySpec` is a Lombok `@Builder` POJO; `WorkerNetworkPolicySpec.builder().metricsPort(9404).metricsIngress(List.of(scraper)).build()` removes the getter stubs in all four tests, the `Mockito defaults Integer to 0` comment, and the redundant `thenReturn(null)` added to `setUp` (that is already Mockito's default). `KueueWorkloadFactoryTest` uses `WorkerSpec.builder()` the same way.

dongjoon-hyun (commented):
Thank you for the update, @yalindogusahin. The `networkPolicy` nesting, the `SPARK_DAEMON_JAVA_OPTS` switch, the rebase, and the e2e suite all landed as asked. I checked out the branch locally: `:spark-operator-api:assertGeneratedCRDMatchesHelmChart`, `:spark-submission-worker:test` (17 tests), `spotlessCheck`, `checkstyle`, and `javadoc` all pass, and the branch merges cleanly onto current `main`.

There are still a few things to sort out before this can go in. Inline comments carry the details; the summary is below.

## Blockers

1. **The `8081` guard does not guard the web UI port.** The worker UI port is configurable: this constructor forwards every `sparkConf` entry as `-Dkey=value` into `SPARK_WORKER_OPTS`, and Spark's `WorkerArguments` honors `spark.worker.ui.port` (and `SPARK_WORKER_WEBUI_PORT`). So `sparkConf: {spark.worker.ui.port: "9404"}` + `metricsPort: 9404` passes the check and exposes the full UI to the listed peers, which is exactly what the Javadoc and the docs say cannot happen. I would rather take the other option I mentioned earlier: drop the throw and soften the wording ("should be a dedicated exporter port, not the worker UI port; the operator does not verify this"). If you keep a check, it has to be derived from the conf, not a literal.

2. **Validation is at the wrong layer.** An `IllegalArgumentException` from the `SparkClusterResourceSpec` constructor is caught by `ClusterInitStep` as a generic `SchedulingFailure` (an infrastructure-failure state) with a stack trace, and the next reconcile turns it into `Failed` / "Cannot process cluster status" via `ClusterUnknownStateStep`, which immediate-requeues forever and appends a `Failed` entry to `stateTransitionHistory` on every pass. If any validation stays, it belongs in `ClusterValidateStep` (see `AppValidateStep.validateGatewayRoutes` for the pattern) or, better, in the CRD schema: `@Required` on both fields and `@Min(1) @Max(65535)` on `metricsPort` (fabric8 `generator-annotations` is already on the api module's classpath, and `@Required` is used on `RuntimeVersions.sparkVersion`). That also removes the silent no-op when only one of the two fields is set, and lets the null-guard, the "both required" paragraph, and two of the new unit tests go away.

3. **The new e2e group never runs.** `.github/workflows/build_and_test.yml` lists `test-group`s explicitly and `worker-metrics-network-policy` is not among them. The cheapest fix is to not add a group at all: put the `workerSpec.networkPolicy` block into `tests/e2e/state-transition/spark-cluster-example-succeeded.yaml` and append the new assertion as a third `assert:` step there. That suite already runs in the static/dynamic/selector modes.

## Example / docs

- The `spark_worker_*` rule in the example matches nothing: Spark registers only the servlet sink by default, so the `metrics` JMX domain is empty unless `spark.metrics.conf.*.sink.jmx.class=org.apache.spark.metrics.sink.JmxSink` is set. Neither the example `sparkConf` nor the docs mention it, so a user following the walk-through gets `jvm_*` series only.
- The docs step-2 ConfigMap and the example ship different exporter configs (catch-all `.*` vs. targeted rules; `readOnly` differs on the mount). I would keep the full CR only in the example and show just the `networkPolicy` fragment in the docs.

## Minor

- Please update the PR title and description; both still describe the original 8081/`web`-port design that the code now explicitly rejects.
- `WorkerNetworkPolicySpec` is new, so `@since 1.1.0`, not `0.1.0`.
- The new tests mock a Lombok `@Builder` POJO; `WorkerNetworkPolicySpec.builder()...build()` removes the getter stubs, the Mockito-`Integer`-defaults comment, and the redundant `thenReturn(null)` in `setUp`.
- `builder = builder.addNewIngress()...endIngress()` is a no-op reassignment; fabric8's `endIngress()` mutates and returns the same enclosing fluent.

spark-operator-api/src/main/java/org/apache/spark/k8s/operator/spec/WorkerNetworkPolicySpec.java
yalindogusahin: Sorry about this one, that was my mistake I should have checked because the model hallucinated for the version.

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkClusterResourceSpec.java
yalindogusahin: Removed it and updated the docs/Javadoc to recommend a dedicated exporter port and clarify that the operator doesn't verify this.

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkClusterResourceSpec.java
yalindogusahin: Moved validation into the CRD schema. Both fields are now required when the block is present, and the port is constrained to 1–65535. Removed the constructor exception.

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkClusterResourceSpec.java
yalindogusahin: Removed the reassignment and added an early return when the policy is absent or the peer list is empty.

spark-operator-api/src/main/java/org/apache/spark/k8s/operator/spec/WorkerNetworkPolicySpec.java
yalindogusahin: Added the annotations and updated the Helm CRD. Partial blocks are now rejected by the API server. Kept coverage for an explicitly empty peer list, which still adds no rule.

examples/cluster-with-jmx-exporter.yaml
yalindogusahin: Added spark.metrics.conf.*.sink.jmx.class to the example and explained why it's needed in the docs.

docs/operations.md
yalindogusahin: Simplified the docs to show the network policy fragment and link to the full example. Also removed the catch-all exporter rule.

spark-submission-worker/src/test/java/org/apache/spark/k8s/operator/SparkClusterResourceSpecTest.java
yalindogusahin: Switched the tests to WorkerNetworkPolicySpec.builder() and removed the redundant stubs.

yalindogusahin:
Thanks for the detailed review. Pushed the fixes in 0f96579, including moving the policy assertion into the existing state-transition suite and updating the PR title and description. The full build, Javadoc, formatting, Helm lint, and schema checks pass. I haven't run the e2e suite locally because Chainsaw isn't installed.

examples/cluster-with-jmx-exporter.yaml
```diff
@@ -0,0 +1,78 @@
...
```
dongjoon-hyun: This rule still never matches, and my previous comment was incomplete: `JmxSink` alone does not fix it. Spark bundles Dropwizard `metrics-jmx` 4.2.37, whose `DefaultObjectNameFactory` puts both `name` and `type` into the ObjectName, e.g. `metrics:name=worker.coresFree,type=gauges`. jmx_exporter renders that as `metrics<name=worker.coresFree, type=gauges><>Value` and wraps the rule as `^.*(?:pattern).*$`, so `(\S+)>` cannot get past the `, `. I checked the regex against both property orders on a local JVM and it is false for both. With a non-empty `rules` list unmatched beans are dropped, so the scrape has no `spark_worker_*` series at all.

Suggested:

```yaml
    - pattern: 'metrics<name=worker\.(\w+), type=gauges><>Value'
      name: spark_worker_$1
      type: GAUGE
```

A plain `(\S+), type=gauges` would produce `spark_worker_worker_coresFree`, because `WorkerSource.sourceName` is already `worker`. The sentence in `docs/operations.md` ("Without this sink, the exporter only exposes JVM metrics") is right, but please make sure the example actually delivers the other half.

examples/cluster-with-jmx-exporter.yaml
```diff
@@ -0,0 +1,78 @@
...
+kind: ConfigMap
+metadata:
+  name: jmx-exporter-config
+data:
+  jmx-exporter-config.yaml: |
+    rules:
+    - pattern: 'java.lang<type=Memory><HeapMemoryUsage>(\w+)'
+      name: jvm_memory_heap_$1
+      type: GAUGE
+    - pattern: 'java.lang<type=GarbageCollector, name=(\w+)><>CollectionCount'
```
dongjoon-hyun: `name=(\w+)` does not match the default collectors. On a JVM with `-XX:+UseG1GC` the beans are `G1 Young Generation`, `G1 Concurrent GC`, and `G1 Old Generation`; the spaces make this rule fail, and it only matches SerialGC names like `Copy`. G1 is the default on the JDK 17/21 `apache/spark` images once the pod has 2+ CPUs and about 1.8 GB, so `jvm_gc_collection_count` is never emitted. Either use `name=([^>]+)`, or drop the two JVM rules altogether: the javaagent already exports `jvm_memory_*` and `jvm_gc_*` from its built-in collectors.

examples/cluster-with-jmx-exporter.yaml
```diff
@@ -0,0 +1,78 @@
...
+            volumeMounts:
+            - name: jmx-exporter-config
+              mountPath: /etc/metrics
+              readOnly: true
+          volumes:
+          - name: jmx-exporter-config
+            configMap:
+              name: jmx-exporter-config
+  sparkConf:
+    spark.metrics.conf.*.sink.jmx.class: "org.apache.spark.metrics.sink.JmxSink"
```
dongjoon-hyun: Two small things on the example config:

- `sparkConf` entries are forwarded as `-D` to both `SPARK_MASTER_OPTS` and `SPARK_WORKER_OPTS`, so the `*` instance also starts `JmxSink` for the master's `master` and `applications` metrics systems, where no agent reads the MBeans. `spark.metrics.conf.worker.sink.jmx.class` is enough.
- Rules only filter output. Without `includeObjectNames` the agent queries every MBean on every scrape and then discards almost all of it. Something like `includeObjectNames: ["metrics:type=gauges,*"]` (plus the `java.lang` entries if you keep the JVM rules) keeps the scrape cheap.

Also, `docs/spark_custom_resources.md` already documents a `JmxSink` recipe via a mounted `metrics.properties`. A cross-link from the new section would avoid having two unconnected recipes.

spark-operator-api/src/main/java/org/apache/spark/k8s/operator/spec/WorkerNetworkPolicySpec.java
```diff
@@ -0,0 +1,62 @@
...
+  @Max(65535)
+  protected Integer metricsPort;
+
+  /**
+   * Sources allowed to scrape {@link #metricsPort} on the worker pods, in addition to the sources
+   * the generated worker NetworkPolicy admits by default. When this list is empty, no metrics
+   * ingress rule is generated.
+   */
+  @Required
+  protected List<NetworkPolicyPeer> metricsIngress;
```
dongjoon-hyun, line 62: `{metricsPort: 9404, metricsIngress: []}` still passes the schema and silently does nothing, which is the same silent no-op we removed for partial blocks. In the Helm chart an empty list means "deny" around a metrics rule that always exists; here the block's only purpose is to add a rule, so an empty list has no meaning. `@Size(min = 1)` (it is in the same `generator-annotations` jar) emits `minItems: 1`, so the block is either effective or rejected at apply time. Please keep the Java `isEmpty()` guard regardless, see the comment there.

spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkClusterResourceSpec.java
```diff
@@ -467,35 +469,53 @@ private static Optional<PodDisruptionBudget> buildPodDisruptionBudget(
...
+            .endPodSelector()
+            .endFrom()
+            .addNewFrom()
+            .withNewPodSelector()
+            .addToMatchLabels(LABEL_SPARK_ROLE_NAME, LABEL_SPARK_ROLE_DRIVER_VALUE)
+            .endPodSelector()
+            .endFrom()
+            .endIngress();
+    WorkerNetworkPolicySpec networkPolicy = workerSpec.getNetworkPolicy();
+    if (networkPolicy == null || networkPolicy.getMetricsIngress().isEmpty()) {
```
dongjoon-hyun: Optional. I agree nulls cannot arrive through the API server: an old CRD or a `v1beta1` write prunes the whole block, and an explicit `null` is dropped and then fails `required`. So I am not asking for the null checks back. Two notes though:

- This guard is security-relevant, not cosmetic: an ingress rule with an empty or absent `from` admits **all** sources, and fabric8 omits an empty list when serializing. A one-line comment saying so would stop a future cleanup from removing it on the grounds that the schema guarantees a non-empty list.
- For Java callers, `new IntOrString((Integer) null)` is accepted by fabric8 and yields `ports: [{protocol: TCP}]`, i.e. every TCP port for the listed peers. That is fail-open, and nothing pins the `required` list today, so it would become reachable if `@Required` were ever dropped.

docs/operations.md
```diff
@@ -174,6 +174,55 @@ Note that this requires a CNI plugin that enforces NetworkPolicy; on clusters wi
...
+    networkPolicy:
+      metricsPort: 9404
+      metricsIngress:
+        - namespaceSelector:
+            matchLabels:
+              kubernetes.io/metadata.name: "monitoring"
+```
+
+The `networkPolicy` block is optional. When present, the API server requires both fields and a
+port between 1 and 65535. An empty `metricsIngress` list adds no rule. Choose a dedicated exporter
```
dongjoon-hyun: Please add one sentence that the policy is generated when the cluster is created. The `NetworkPolicy` is applied only in `ClusterInitStep` (state `Submitted`) and nothing is reconciled once the cluster is `RunningHealthy`, so a `kubectl edit` that adds or removes a peer is accepted by the API server and then never takes effect. That is existing behavior for every `SparkCluster` field, but scraper peers are the one knob people will edit on a long-running cluster. Changing them means recreating the `SparkCluster`.

dongjoon-hyun (commented):
Thank you for the quick turnaround, @yalindogusahin. All ten points from the previous round are addressed, and the new title and description match the code.

I resolved the `docs/operations.md` conflict locally on top of current `main` and ran the checks there: `gradle build -x test`, `:spark-operator-api:assertGeneratedCRDMatchesHelmChart`, `SparkClusterResourceSpecTest`, the `spark-operator-api` tests, and `helm lint --strict` all pass. `minimum: 1.0` / `maximum: 65535.0` on an integer field is standard fabric8 output and fine. I have not run the Chainsaw suite either, so CI will be the first real run of the new assertion.

A few things are left. Inline comments carry the details.

## Needed before merge

1. **Please rebase.** The branch is 24 commits behind `main` and conflicts in `docs/operations.md`: SPARK-59504 appended a deprecation paragraph for `operatorDeployment.networkPolicy.enable` exactly where this PR inserts the new section. The resolution is mechanical: keep that paragraph at the end of "Restricting Network Access to the Operator", then the new "Exposing SparkCluster Worker Metrics" section.

2. **The example still exports no Spark metrics.** My previous comment was incomplete here, sorry: enabling `JmxSink` is necessary but not sufficient. Neither the `spark_worker_*` rule nor the GC rule matches the MBean names a real JVM produces (see the two inline comments; I checked both against a local JVM).

## Design

3. **Invalid peers now fail late, after the workers are running.** This is in `ClusterInitStep`, which the PR does not touch, so no inline comment. Until now the worker `NetworkPolicy` was static and could not be rejected. It is now built from user input, and the generated peer schema is unconstrained: `metricsIngress: [{}]`, `ipBlock: {}` without `cidr`, or `ipBlock` combined with a selector all pass the CRD and are rejected only by the NetworkPolicy API at server-side apply. `ClusterInitStep` applies the policy after both Services and both StatefulSets, so a typo in a peer leaves the cluster in `SchedulingFailure` / `Failed` with worker pods up and no `NetworkPolicy` at all, which is worse than the unconfigured default. Moving the `NetworkPolicy` apply ahead of the two StatefulSets in `ClusterInitStep` is a small reorder and means a rejected policy creates no pods. I think it belongs in this PR since this PR is what makes the policy fallible.

## Minor

- The description says "Twelve local JSON Schema checks cover omitted and partial blocks, null peers, port boundaries, and a namespace selector", but those checks are not in the diff or in the listed commands, so nobody can reproduce them. Please either add the commands or drop the sentence.

yalindogusahin:
Thanks for the detailed review. I rebased and pushed 79d6571.

The example now matches actual worker gauge MBeans, uses a worker-only JmxSink, and limits JMX queries. With the exporter jar supplied, I scraped a local Spark 4.1.2 worker and saw spark_worker_coresFree, spark_worker_coresUsed, spark_worker_executors, spark_worker_memFree_MB, and spark_worker_memUsed_MB.

I also made metricsIngress nonempty in the CRD, kept the Java policy builder fail closed, moved NetworkPolicy creation before the StatefulSets, and documented that changing scraper peers requires recreating the cluster. I updated the PR description to reflect these changes.

spark-operator-api/src/main/java/org/apache/spark/k8s/operator/spec/WorkerSpec.java
```diff
@@ -45,4 +45,10 @@ public class WorkerSpec {
   protected ServiceSpec serviceSpec;
   protected ObjectMeta serviceMetadata;
   protected HorizontalPodAutoscalerSpec horizontalPodAutoscalerSpec;
+
+  /**
+   * Network policy controlling ingress to the worker pods. When null, the generated worker
+   * NetworkPolicy stays as restrictive as it is without these fields.
+   */
```
dongjoon-hyun: Shall we remove these lines like the other field?

```suggestion
```

docs/operations.md
```diff
@@ -205,6 +205,58 @@ It is still honored: the NetworkPolicy is created when either key is `true`, so
...
+
+Use the [Prometheus JMX Exporter](https://github.com/prometheus/jmx_exporter)
+(`jmx_prometheus_javaagent`) to serve metrics on a dedicated HTTP port. See
+[examples/cluster-with-jmx-exporter.yaml](../examples/cluster-with-jmx-exporter.yaml) for the full
+`ConfigMap` and `SparkCluster` configuration. Replace the example's image with a custom Spark image
+containing the exporter jar at `/opt/jmx_exporter/jmx_prometheus_javaagent.jar`; the stock
+`apache/spark` image does not include it.
+
+The example mounts the exporter rules and attaches the agent through `SPARK_DAEMON_JAVA_OPTS`.
+The operator sets `SPARK_WORKER_OPTS` itself, so a value there would be overwritten. It also sets
```
dongjoon-hyun, line 224: This describes the wrong failure mode. The operator does not replace a user-provided `SPARK_WORKER_OPTS`; it appends a second env entry with the same name (`addNewEnv()` in `SparkClusterResourceSpec.java:340`), and `ClusterInitStep` applies the StatefulSet with `serverSideApply()`, which rejects duplicate keys in the `env` associative list.

Checked against a live API server:

```
$ kubectl apply --server-side --dry-run=server -f dup-env.yaml
Error from server: failed to create typed patch object (default/dup-env-probe; apps/v1, Kind=StatefulSet):
  .spec.template.spec.containers[name="worker"].env: duplicate entries for key [name="SPARK_WORKER_OPTS"]
```

So a reader who takes "would be overwritten" at face value and sets it anyway does not get a silently ignored value. The whole worker StatefulSet apply is rejected, the cluster goes to `SchedulingFailure` and then `Failed`, and the status message is a stack trace that never mentions the duplicate variable.

The advice to use `SPARK_DAEMON_JAVA_OPTS` is right, only the reason is wrong. Something like: "`SPARK_WORKER_OPTS` must not be set there: the operator appends its own, and server-side apply rejects duplicate env names."

dongjoon-hyun (commented):
Thanks, @yalindogusahin. This is close now. Seven of the nine items from the last round are resolved, and the two that remain are documentation-only.

What I verified locally:

- `gradle build` and `:spark-operator-api:assertGeneratedCRDMatchesHelmChart` pass, 347 unit tests green, `helm lint --strict` clean, and the branch merges cleanly again after the rebase.
- The new exporter rule matches. I rendered the five `WorkerSource` gauges the way Dropwizard's `DefaultObjectNameFactory` names them and ran your pattern against them on a JVM: all five match and produce exactly `spark_worker_coresFree`, `coresUsed`, `executors`, `memFree_MB`, `memUsed_MB`, which is what you saw when you scraped. `includeObjectNames: ["metrics:type=gauges,*"]` is a valid ObjectName pattern and selects gauges only.
- The generated schema is accepted by a real API server. I dry-run applied the new fragment (`minimum: 1.0` / `maximum: 65535.0` on `type: integer`, `minItems: 1`, nested `required`) against a 1.37 server and it went through, so the float bounds fabric8 emits are fine.
- The `ClusterInitStep` reorder looks right. Master pods carry the cluster-name label, so the existing rule still admits master traffic to workers; the suspend path returns before the try block and is unaffected; and the new test pins that no StatefulSet is created when the policy is rejected. As a bonus, worker pods are now never briefly unprotected.
- No cross-suite fallout: `spark-cluster-example-succeeded.yaml` is used only by `state-transition`, and yours is the only NetworkPolicy assertion under `tests/e2e`.

Two small things left, both inline, neither affecting the design.

examples/cluster-with-jmx-exporter.yaml
```diff
@@ -0,0 +1,71 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: jmx-exporter-config
+data:
+  jmx-exporter-config.yaml: |
+    includeObjectNames: ["metrics:type=gauges,*"]
```
dongjoon-hyun, line 21: Minor: the example asks the user to bake their own `jmx_prometheus_javaagent` jar but never names a version, and `includeObjectNames` is the newer key name. Older agent releases used `whitelistObjectNames` and reject unrecognized config keys outright. If someone picks an old jar, the agent fails to load this config, `-javaagent` initialization fails and the worker JVM aborts, so the pods CrashLoopBackOff while the `SparkCluster` still reports `RunningHealthy` (the operator sets that as soon as the applies succeed).

A minimum version next to the jar path comment would close this, e.g. "requires jmx_prometheus_javaagent 1.x; the `includeObjectNames` key is not accepted by older releases". I confirmed the key and the pattern work, but not which release introduced `includeObjectNames`, so please use whatever floor you are confident in.

PR #795, opened 2026-08-18
[SPARK-58835] Enable `spark.shuffle.service.enabled` by default in `SparkCluster`

dongjoon-hyun:

This PR aims to enable `spark.shuffle.service.enabled` by default in `SparkCluster` resources. The default is applied before the user-provided `sparkConf` and `confOverrides`, so any user-specified value takes precedence.

In Standalone mode, dynamic allocation requires the external shuffle service, which must be enabled at Worker launch time. Defaulting it to `true` makes every `SparkCluster` dynamic-allocation ready out of the box, while users can still opt out by setting `spark.shuffle.service.enabled=false` in `spec.sparkConf`.

Yes. `SparkCluster` Workers now start the external shuffle service by default. Setting `spark.shuffle.service.enabled=false` in `spec.sparkConf` restores the previous behavior.

Pass the CIs with two new test cases in `SparkClusterSubmissionWorkerTest` covering the default and the user-override case.

Generated-by: Claude Fable 5


peter-toth (approved):

dongjoon-hyun:

dongjoon-hyun:
Merged to main

PR #800, opened 2026-08-23
[SPARK-58958] Remove stale entries from `verification-metadata.xml`

dongjoon-hyun:

This PR regenerates `gradle/verification-metadata.xml` from scratch to drop stale entries:

```bash
rm gradle/verification-metadata.xml
gradle --write-verification-metadata sha512 build --refresh-dependencies
```

Since `--write-verification-metadata` is append-only, the checked-in file accumulated entries that are no longer recorded. The regeneration removes 31 artifact entries (987 → 956) and adds none:

- 30 `.pom` entries of components that also publish Gradle Module Metadata. Gradle uses the `.module` file as the descriptor for those components, so the sibling `.pom` is never verified. Each component keeps its `.module` entry.
- 1 `junit-bom-6.0.2.module` entry, which is genuinely stale since only `junit-bom` 6.0.3 / 6.1.x are resolved today.

No `.jar` checksum is removed and no dependency version is changed.

To keep the dependency verification metadata in sync with the actual dependency set, so that leftovers don't show up as noise in future dependency upgrade diffs.

No.

Pass the CIs and `gradle build` locally with `verify-metadata=true` enabled. The pruned file was also validated against an empty `GRADLE_USER_HOME` to force a cold re-download of all dependencies.

Generated-by: Claude Opus 5


peter-toth (approved):

dongjoon-hyun:

dongjoon-hyun:
Merged to main

PR #806, opened 2026-08-28
[SPARK-59092] Honor Retry-After when retrying secondary resource creation on Kubernetes API errors

jiangzho:

When ReconcilerUtils.getOrCreateSecondaryResource retries a failed secondary-resource creation, it now honors the Kubernetes API server's requested retry delay instead of always falling back to our own exponential backoff:

- BackoffUtils.getRetryAfterMillis reads KubernetesClientException.getStatus().getDetails().getRetryAfterSeconds() — the field the API server mirrors from the HTTP Retry-After header into the response body (fabric8's client doesn't expose the raw header itself). When present and positive, BackoffUtils.backoffSleep sleeps for that duration instead of computing a jittered exponential backoff.
- ReconcilerUtils.shouldBackoffBeforeRetry now also returns true whenever a Retry-After value is present, regardless of status code — previously only 409 and 429 triggered a backoff sleep, so a transient 503 carrying retryAfterSeconds was retried immediately, ignoring the server's hint.
 - The server-provided delay is capped at `spark.kubernetes.operator.api.secondaryResourceCreateMaxBackoffMillis` (BackoffUtils.computeDelayMillis), so a server or intermediate proxy returning an excessive Retry-After (e.g. an hour) cannot stall the reconciler thread for an unbounded amount of time.
 - The configured jitter (`spark.kubernetes.operator.api.secondaryResourceCreateBackoffJitterMillis`) is added on top of the capped Retry-After delay as well, not just the exponential-backoff path
- Simplified backoffSleep: it now evaluates getRetryAfterMillis(e) once and logs a single line reporting both the server-requested delay and the actual (capped/jittered) delay, so capping/jittering is visible when diagnosing retry behavior, instead of two near-duplicate log.info calls.
- Corrected the Javadoc/description(...) text for `API_SECONDARY_RESOURCE_CREATE_MAX_ATTEMPTS` and `API_SECONDARY_RESOURCE_CREATE_INITIAL_BACKOFF_MILLIS` in SparkOperatorConf.java, which previously stated backoff only applies "on 409 (Conflict) and 429 (Too Many Requests) responses" — now describing the actual Retry-After-driven behavior. 

 The server is in the best position to say how long to wait before retrying a throttled or overloaded request. Always using our own backoff schedule can retry too soon (extending throttling) or wait longer than necessary, and previously we ignored the hint entirely for non-429/409 errors like 503. The delay is still capped and jittered to bound worst-case wait time and avoid synchronized retry bursts across concurrently throttled reconciles, and the config docs are corrected to accurately describe when backoff applies.

###  Does this PR introduce any user-facing change?

No behavior change for existing 409/429 handling beyond jitter. Retryable errors that previously retried immediately (e.g. 503) will now wait if the server supplies a Retry-After hint, capped and jittered per the existing backoff config. Two config property descriptions were corrected for accuracy; no config keys, defaults, or semantics changed. Log line format for backoff retries changed (single consolidated log.info message with requested + actual delay).

- BackoffUtilsTest: covers getRetryAfterMillis (present/absent/non-positive/no-status cases); computeDelayMillis uses the Retry-After value plus jitter when below max backoff, caps at API_SECONDARY_RESOURCE_CREATE_MAX_BACKOFF_MILLIS plus jitter when the server-provided value is excessive, and falls back to exponential backoff when Retry-After is absent; plus existing computeBackoffDelay bounds checks.
- ReconcilerUtilsTest: retriesCreateOn429AndSucceedsHonoringRetryAfter (429 with retryAfterSeconds=1) and retriesCreateOnTransient5xxHonoringRetryAfterWhenPresent (503 with retryAfterSeconds=1), both asserting the retry succeeds and waits at least 1s (lower-bound assertion, compatible with jitter).
- Verified docs/config_properties.md is up to date via `./gradlew build-tools-docs-utils:generateConfPropsDoc -Poverwrite` (keeps validateConfPropsDocUpToDate passing).

###  Was this patch authored or co-authored using generative AI tooling?

 Co-Authored-by: Claude Code (claude-sonnet-5)


dongjoon-hyun:
Thank you for working on this, @jiangzho . The direction looks good to me — honoring the server-provided `retryAfterSeconds` is the right thing to do. I have one blocker and one design concern.

**1. `retriesCreateOnTransient5xxHonoringRetryAfterWhenPresent` actually fails (blocker)**

I ran the PR branch locally and this test fails at the timing assertion:

```
ReconcilerUtilsTest > retriesCreateOnTransient5xxHonoringRetryAfterWhenPresent() FAILED
    org.opentest4j.AssertionFailedError at ReconcilerUtilsTest.java:189
```

The test never reaches the backoff sleep. Since 503 is a transient error, the catch block in `getOrCreateSecondaryResource` performs the GET fallback **before** the backoff sleep, and the test stubs the second `get()` to return the pod — so the method returns immediately from inside the catch block (elapsed ≈ 0ms < 1000ms). To exercise the intended path (503 + Retry-After → sleep → retry), the second `get()` should return `null` and the second `create()` should succeed.

**2. The server-provided delay is not capped**

`getRetryAfterMillis` feeds `Thread.sleep` directly. The existing exponential backoff is capped by `API_SECONDARY_RESOURCE_CREATE_MAX_BACKOFF_MILLIS`, but the server-provided value is trusted without a bound. Since this sleep runs on the reconciler thread, an API server (or an intermediate proxy) responding with e.g. `retryAfterSeconds: 3600` would block reconciliation for an hour. I'd suggest applying the existing cap, e.g. `Math.min(retryAfterMillis, maxBackoffMs)`.

**Minor**: the two new `ReconcilerUtilsTest` tests each sleep a real 1s, adding ~2s to the unit test suite. Probably unavoidable given `retryAfterSeconds` has whole-second granularity, but worth noting.

dongjoon-hyun:
BTW, it seems that this PR is under active development via force-pushing . In that case, please use `Draft` status in the community, @jiangzho .

dongjoon-hyun:
I converted this PR to `Draft` status. Please convert this back to normal when the PR is ready for review.

jiangzho:
Thanks @dongjoon-hyun for the review!

Sorry I was working on the flaky test and did a few force push, it should not be ready for review.
I also added cap as suggested so a server or intermediate proxy can't stall the reconciler thread beyond our configured ceiling.

dongjoon-hyun (approved):
Thank you for the update, @jiangzho . Both of my earlier concerns are addressed: the 503 test now keeps `get()` returning `null` so it actually exercises the backoff sleep path, and the server-provided delay is capped via `Math.min(retryAfterMillis, maxBackoffMs)` with `computeDelayMillisCapsHugeRetryAfterAtConfiguredMaxBackoff` covering it.

I ran the branch locally and it is green:

```
gradle spotlessCheck checkstyleMain :spark-operator:test --tests "org.apache.spark.k8s.operator.utils.*"
BUILD SUCCESSFUL
```

`BackoffUtilsTest` (8 tests) and `ReconcilerUtilsTest` (13 tests) all pass.

Approving. Please address (1) before merging, the rest are optional.

**1. Stale config descriptions and generated docs (will fail CI)**

The behavior changed from "backoff only on 409/429" to "backoff whenever the server sends a Retry-After", but the config descriptions still state the old rule:

- `SparkOperatorConf.java` — `API_SECONDARY_RESOURCE_CREATE_MAX_ATTEMPTS`: "Exponential backoff with jitter is applied before retrying on 409 (Conflict) and 429 (Too Many Requests) responses. Other retryable errors are retried immediately."
- `SparkOperatorConf.java` — `API_SECONDARY_RESOURCE_CREATE_INITIAL_BACKOFF_MILLIS`: "Backoff is applied only on 409 (Conflict) and 429 (Too Many Requests) responses."

`docs/config_properties.md` is generated from these, so it needs to be regenerated after the wording is fixed, otherwise `validateConfPropsDocUpToDate` fails:

```
gradle build-tools-docs-utils:generateConfPropsDoc -Poverwrite
```

**2. PR description is out of sync with the code**

"When present and positive, `BackoffUtils.backoffSleep` sleeps for exactly that duration" — now that the cap is in, please mention it is capped at `secondaryResourceCreateMaxBackoffMillis`.

**3. No jitter on the Retry-After path**

`computeDelayMillis` only adds jitter on the exponential path. When many secondary resource creations are throttled by API Priority and Fairness at the same time, they all receive the same `retryAfterSeconds` and retry at exactly the same instant. client-go uses Retry-After as-is, but since we already have a configured jitter, reusing it here would avoid the synchronized retry burst, e.g. `Math.min(retryAfterMillis, maxBackoffMs) + jitter`.

**4. Simplify the log branch in `backoffSleep`**

The two `log.info` calls differ only in the leading phrase and take the same four arguments, and `getRetryAfterMillis(e)` is evaluated twice (once inside `computeDelayMillis`, once for the branch). A single statement keeps the diff smaller:

```java
  public static void backoffSleep(
      KubernetesClientException e, long attemptCount, long maxAttempts) {
    long actualDelay = computeDelayMillis(e, attemptCount);
    log.info(
        "Retrying resource creation with {}. Delay: {}ms, responseCode: {}, attempt: {}/{}",
        getRetryAfterMillis(e) != null ? "server-requested Retry-After" : "exponential backoff",
        actualDelay,
        e.getCode(),
        attemptCount,
        maxAttempts);
```

Relatedly, when the server asks for 3600s and we cap it to 40s, the log still reads "honoring server Retry-After. Delay: 40000ms", which is a bit misleading. Logging the requested value alongside the actual one would help when diagnosing.

**5. Timing-dependent tests (minor, follow-up on my earlier note)**

The two new tests each sleep a real second and assert on wall clock (`elapsedMillis >= 1000L`), pushing `ReconcilerUtilsTest` to ~2.7s. Whole-second granularity makes that hard to avoid as written, but what these tests really pin down is "a Retry-After makes `shouldBackoffBeforeRetry` return true". Making `shouldBackoffBeforeRetry` package-private and asserting on it directly would catch the same regression without sleeping; the delay computation itself is already precisely covered by `BackoffUtilsTest`.

**6. Nit**

Changing the signature from `int responseCode` to `KubernetesClientException` couples the backoff utility to the fabric8 exception type. Harmless in practice since `ReconcilerUtils` is the only caller, but passing a `Long retryAfterMillis` would keep `BackoffUtils` independent.

dongjoon-hyun:
Could you re-trigger the CIs, @jiangzho ?

Screenshot 2026-09-03 at 09 48 59

jiangzho:
Hi Dongjoon - I tried a few times but these are constantly failing with `moby/buildkit:buildx-stable-1` 404. Since I don't have write access to retrigger checks on this repo directly, I'll stop spinning cycles on my fork — could you retrigger the checks here when you get a chance? If it's still failing, it's likely a broader GitHub Actions/Docker Hub issue worth flagging separately.

dongjoon-hyun:
Got it. Let me investigate the failure to unblock you, @jiangzho .

dongjoon-hyun:
I made a PR to unblock this PR. It's a Minikube issue. @jiangzho .
- https://github.com/apache/spark-kubernetes-operator/pull/807

dongjoon-hyun:
The main branch is recovered. Please re-trigger your GitHub Actions CI on your fork. As you see in the following, the community member cannot re-trigger your CIs because it's owned by only you.

Screenshot 2026-09-03 at 12 49 03

dongjoon-hyun:
It seems that another CI outage is caused by ASF Infra team at `Gradle` GitHub Actions pattern. I'm digging it now.
- https://github.com/apache/infrastructure-actions/commit/63206b1

dongjoon-hyun:
I made another PR to comply ASF Infra policy change.
- https://github.com/apache/spark-kubernetes-operator/pull/808

jiangzho:
The build workflow is happy new after merging from main, it just not reports back the succeeded results

dongjoon-hyun:
Merged to main

dongjoon-hyun:
Thank you, @jiangzho !

PR #811, opened 2026-09-04
[SPARK-59262] Add Java 27-ea to GitHub Actions

dongjoon-hyun:

This PR aims to add Java `27-ea` to the `build-test` GitHub Actions matrix as the runner environment.

Apache Spark K8s Operator uses a `Java 26` toolchain. Like [SPARK-55737] which added `26-ea` before its GA, this PR only aims to test `Java 27-ea` as the runner environment ahead of the [JDK 27 GA on 2026-09-15](https://openjdk.org/projects/jdk/27/).

Note that `-XX:+UseCompactObjectHeaders` is intentionally left as is, because [JEP 534: Compact Object Headers by Default](https://openjdk.org/jeps/534) makes it the default in Java 27. Passing the flag explicitly would be a no-op there, while Java 25 and 26 still require it.

No.

Pass the CIs and check the log.

Generated-by: Claude Opus 5

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -29,7 +29,7 @@ jobs:
       max-parallel: 20
       matrix:
         os: [ 'ubuntu-26.04', 'ubuntu-26.04-arm' ]
-        java-version: [ 21, 25, 26 ]
+        java-version: [ 21, 25, 26, '27-ea' ]
     steps:
       - name: Checkout repository
         uses: actions/checkout@v6


peter-toth (approved):

dongjoon-hyun:
Thank you, @peter-toth !

dongjoon-hyun:
Merged to main

dongjoon-hyun:
Oh, although all CIs passed on this PR due to the cache-hit, but Gradle seems to complain when it needs to build from the scratch actually.

Screenshot 2026-09-04 at 12 56 02

dongjoon-hyun:
Let me revert this for now until we have Gradle 9.8.

dongjoon-hyun:
This is reverted via c910374

PR #833, opened 2026-09-16
[SPARK-59586] Support `spark.kubernetes.operator.events.excludedReasons` to skip K8s events by reason

dongjoon-hyun:

This PR aims to support `spark.kubernetes.operator.events.excludedReasons`, a comma-separated list of Java regular expressions. Events whose reason fully matches any of them are not published.

```properties
spark.kubernetes.operator.events.excludedReasons=RunningHealthy,RunningWithPartialCapacity
spark.kubernetes.operator.events.excludedReasons=Running.*
```

Since SPARK-59576, the operator publishes K8s events for all state transitions. Some events, such as `RunningHealthy`, can be noisy for long-running resources.
- https://github.com/apache/spark-kubernetes-operator/pull/831

No behavior change by default because the default value is empty.

Pass the CIs with the newly added test cases.

Generated-by: Claude Opus 5


dongjoon-hyun:
cc @TQJADE

spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/ConfigurableEventRecorder.java
```diff
@@ -94,6 +102,24 @@ private static boolean eventsEnabled() {
     return Boolean.TRUE.equals(KUBERNETES_EVENTS_ENABLED.getValue());
   }
 
+  private static boolean isExcluded(String reason) {
+    // A null resolved value is treated as an empty list.
+    return Utils.sanitizeCommaSeparatedStrAsSet(KUBERNETES_EVENTS_EXCLUDED_REASONS.getValue())
```
peter-toth: **Finding 1.** This helper carries a sentinel that belongs to a different option. `Utils.java:71-73`:

```java
    if ("*".equals(str)) {
      return Set.of();
    }
```

That is right for `OPERATOR_WATCHED_NAMESPACES` (`Utils.java:202`), where `*` means "watch everything, so do not restrict". Here the same branch empties the exclusion list, so the whole value `*` excludes nothing. I confirmed both halves with a probe test:

```java
  setExcludedReasons("*");
  recorder.record(EventRecord.normal("RunningHealthy", "published?"), context);
  verify(delegate).record(event, context);        // passes: * excludes nothing

  setExcludedReasons(".*");
  recorder.record(EventRecord.normal("RunningHealthy", "dropped?"), context);
  verifyNoInteractions(delegate);                 // passes: .* excludes everything
```

The end state is not wrong against the docs - `*` is not a valid Java regex, and the description already says an invalid expression "matches only the reason identical to it", which also excludes nothing. What is lost is the signal: the sentinel returns before `matches` ever runs, so `*` is the one malformed value that produces no warning, and a user who reaches for the obvious wildcard gets silence. It also leaves this option's semantics dependent on a helper whose `*` case exists for another feature.

Splitting locally keeps the documented rule as the only rule:

```java
  private static boolean isExcluded(String reason) {
    // A null or blank resolved value is treated as an empty list. The comma splitting is local
    // rather than Utils.sanitizeCommaSeparatedStrAsSet, whose "*" sentinel means "no restriction"
    // for the watched-namespaces option and would silently empty this list instead.
    String value = KUBERNETES_EVENTS_EXCLUDED_REASONS.getValue();
    if (StringUtils.isBlank(value)) {
      return false;
    }
    return Arrays.stream(value.split(","))
        .map(String::trim)
        .filter(StringUtils::isNotBlank)
        .anyMatch(regex -> matches(regex, reason));
  }
```

If you would rather keep the helper, a test pinning `*` plus a sentence next to the `,` caveat at `docs/configuration.md:92` would at least make the behaviour discoverable.

peter-toth (commented):

`ConfigurableEventRecorder.record` is the single funnel every event goes through, so gating on the reason there covers the state transitions from #831 and the `ReconcileError` / `CleanupError` / `StatusUpdateFailed` warnings alike. Both description examples check out against the guard, the whole-match semantics is pinned (I swapped `Pattern.matches` for a `find()` and `excludedReasonRegexMustMatchTheWholeReason` is the one test that fails), and the default stays a no-op. One thing about the edges: `*` is silently swallowed by the shared comma helper.

### Non-blocking
- **1.** *`*` is swallowed by `sanitizeCommaSeparatedStrAsSet` before it can be reported:* that helper maps the whole value `*` to an empty set, a sentinel meaning "no restriction" for its original caller `OPERATOR_WATCHED_NAMESPACES`. So `excludedReasons=*` excludes nothing and logs nothing, where every other invalid expression at least warns. [inline: [`ConfigurableEventRecorder.java:107`](https://github.com/apache/spark-kubernetes-operator/pull/833#discussion_r4029048018)]

peter-toth (approved):

spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/ConfigurableEventRecorder.java
dongjoon-hyun: Thank you, @peter-toth. Addressed in 1b1a923. I switched to local comma splitting so that `*` is reported like other invalid expressions, and added a test and a doc note that `.*` is the way to exclude all reasons.

dongjoon-hyun:
Merged to main

PR #835, opened 2026-09-16
[SPARK-59588] Add `KueueWorkloadUtils` to request admission and release Kueue Workloads

dongjoon-hyun:

This PR aims to add `KueueWorkloadUtils`, the first part of the Kueue runtime integration for `SparkApplication` and `SparkCluster` resources that have the `kueue.x-k8s.io/queue-name` label.

- `requestAdmission` gets or creates a `Workload` and returns one of the following results.
  - `ADMITTED`: Kueue admitted the `Workload`.
  - `PENDING`: The `Workload` is waiting for quota.
  - `STALE`: The `Workload` cannot be used because it is being deleted, belongs to another resource, or is pending with outdated pod sets. In the last two cases, it is deleted without waiting so that it can be recreated.
- `releaseWorkload` deletes the `Workload` to release its quota. If the deletion fails, it only logs a warning.
- `KueueWorkloadFactory.getWorkloadName` becomes package-private.

Nothing calls this class yet. Follow-up PRs will use it in `AppInitStep`, `ClusterInitStep` and `AppCleanUpStep`.

To support Kueue admission control for `SparkApplication` and `SparkCluster` in smaller PRs.

No. Nothing calls the new class yet.

Pass the CIs with the newly added `KueueWorkloadUtilsTest`.

Generated-by: Claude Opus 5


peter-toth (commented):

The three-way `ADMITTED` / `PENDING` / `STALE` split reads well, and hashing the desired pod sets into an annotation is the right way to compare against a stored spec the API server has defaulted. Two things I checked and can confirm: `WorkloadStatus.isAdmitted` is exactly Kueue's own `IsAdmitted` (`pkg/workload/workload.go:1294`), and `SyncAdmittedCondition` flips that condition to `False` when quota reservation is lost, so an evicted Workload correctly stops reading as admitted. The gap is on the test side: every test builds a `PodSet` with no template, so the invariant the `STALE` branch rests on is never exercised on a realistic Workload.

### Non-blocking
- **1.** *`hashPodSets` is never tested on a pod set with a template:* all eight tests use `PodSet.builder().name("executor").count(n)`, so the hash comparison only ever sees a `count` change. If two builds of the same spec ever hashed differently, the pending Workload would be deleted and recreated every 5s and never get admitted. [inline: [`KueueWorkloadUtilsTest.java:208`](https://github.com/apache/spark-kubernetes-operator/pull/835#discussion_r4029146874)]

### Minor
- **2.** *`requestAdmission`'s javadoc documents only the return values:* it can also throw `IllegalStateException` and, from the two `STALE` delete branches, `KubernetesClientException`. [inline: [`KueueWorkloadUtils.java:80`](https://github.com/apache/spark-kubernetes-operator/pull/835#discussion_r4029146888)]

spark-operator/src/test/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtilsTest.java
```diff
@@ -0,0 +1,212 @@
...
+                    .withKind("SparkApplication")
+                    .withUid(ownerUid)
+                    .withController(true)
+                    .build())
+            .build());
+    workload.setSpec(
+        WorkloadSpec.builder()
+            .queueName("test-queue")
+            .active(true)
+            .podSets(List.of(PodSet.builder().name("executor").count(executors).build()))
```
peter-toth, line 266: **Finding 1.** No test gives a `PodSet` a `template`, so `hashPodSets` is only ever exercised on `count`. In production every pod set carries a full `PodTemplateSpec` from `buildPodSet`, and that is where everything non-trivial lives - containers, resource maps, and the node selector that `KueueWorkloadFactory.java:443` builds into a `new HashMap<>()`.

That matters because "the same spec hashes the same on every reconcile" is what keeps the `STALE` branch from livelocking. If it ever stopped holding, a pending Workload would be found outdated, deleted, recreated, found outdated again, once per `STALE_WORKLOAD_REQUEUE_INTERVAL`, and never admitted. Nothing in the suite would notice.

I checked that the invariant does hold today, so this is a coverage gap rather than a bug. The probe was:

```java
  @Test
  void hashIsStableAcrossIdenticalBuilds() {
    assertEquals(
        KueueWorkloadUtils.hashPodSets(KueueWorkloadFactory.buildWorkload(probeApp())),
        KueueWorkloadUtils.hashPodSets(KueueWorkloadFactory.buildWorkload(probeApp())));
  }
```

with `probeApp()` setting `spark.executor.instances` plus twelve `spark.kubernetes.node.selector.*` entries, so both builds go through the `HashMap` node selector. It passes. Something along those lines in `KueueWorkloadFactoryTest`, where the app builders already live, would pin it.

While there, the `.with(SerializationFeature.ORDER_MAP_ENTRIES_BY_KEYS)` on the writer is also unpinned - I removed it and all 34 tests under `org.apache.spark.k8s.operator.kueue` stayed green, including the probe above. It is doing no work today because both sides of the comparison come from the same factory path, but it is the line that would matter if a map ever reached the hash from a different source.

spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java
```diff
@@ -0,0 +1,182 @@
...
+     */
+    STALE
+  }
+
+  /**
+   * Creates the given Workload if it does not exist yet and reports whether Kueue admitted it.
+   *
+   * @param client The KubernetesClient.
+   * @param desired The Workload built for the resource to be admitted.
+   * @return The AdmissionResult for the Workload.
```
peter-toth, line 81: **Finding 2.** The class exists to be called from `AppInitStep`, `ClusterInitStep` and `AppCleanUpStep`, so its throws contract is the part those callers have to design around, and right now it is only in the body:

- `:93` throws `IllegalStateException` when `getOrCreateSecondaryResource` returns empty.
- `:148` `deleteWorkload` does not catch, so the two `STALE` branches can throw `KubernetesClientException`. That is a deliberate difference from `releaseWorkload`, which swallows and explains why in its own javadoc, but nothing says so here.

```java
   * @param desired The Workload built for the resource to be admitted.
   * @return The AdmissionResult for the Workload.
   * @throws IllegalStateException if the Workload can neither be read nor created.
   * @throws io.fabric8.kubernetes.client.KubernetesClientException if a stale Workload cannot be
   *     deleted. Unlike {@link #releaseWorkload}, this is not swallowed: the resource must not be
   *     created while a Workload of the wrong shape still holds the quota.
   */
```

Adjust the reasoning if I have the intent backwards - either way it is worth stating, since the difference between the two delete paths is the kind of thing a caller gets wrong once.

The `desired` argument is also mutated in place at `:70` (the hash annotation is added to the caller's object). Harmless with the current factory, which builds a fresh Workload per call, but worth a `@param` note if this stays public.

peter-toth (approved):

spark-operator/src/test/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtilsTest.java
dongjoon-hyun: Thank you. I added `hashPodSetsIsStableForTheSameSpec` and `hashPodSetsIgnoresMapEntryOrder` to `KueueWorkloadUtilsTest`, next to `hashPodSets`, in 4141188. The first test builds the same app with 12 node selectors twice through `KueueWorkloadFactory.buildWorkload`. The second test fails if `ORDER_MAP_ENTRIES_BY_KEYS` is removed, so that line is now covered too.

spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java
dongjoon-hyun: Thank you. I documented both `@throws` cases and the in-place update of `desired` in the Javadoc of `requestAdmission` in 4141188. Your reading of the intent is correct: a failed stale deletion is not swallowed, so the resource is not created until the stale Workload is gone.

dongjoon-hyun:

dongjoon-hyun:
Merged to main

PR #837, opened 2026-09-16
[SPARK-59594] Release Kueue `Workload` when `AppCleanUpStep` releases `SparkApplication` resources

dongjoon-hyun:

This PR aims to delete the Kueue `Workload` when `AppCleanUpStep` releases the resources of a `SparkApplication` with the `kueue.x-k8s.io/queue-name` label.

The `Workload` is deleted before the attempt's resources, so it is released on every path that reaches the resource release:
- On-demand clean-up
- Stopping apps, including before a restart
- Terminated apps whose retention period has expired
- `SchedulingFailure` attempts whose spec cannot be built again

Apps that keep their resources via `resourceRetainPolicy` keep their quota until the resources are released.

This is part of the Kueue runtime integration. Once labeled apps wait for admission, a `Workload` left `Admitted` after the app finishes holds the `ClusterQueue` quota indefinitely by default (`ttlAfterStopMillis=-1`). This blocks other jobs in the same `LocalQueue`.

- Deleting the `Workload` returns the quota and needs no `workloads/status` permission. A restarted attempt is queued again with a new `Workload`, so the quota is not held during the restart backoff.
- Failures are only logged as warnings, and a missing `Workload` or Kueue CRD is a no-op. So clean-up is never blocked, and `ownerReference` GC still removes the `Workload` eventually.
- This should land before the admission PR so that `main` never leaks quota.

No, the operator does not create Kueue Workloads yet.

Pass the CIs with the newly added test cases and the updated existing test cases in `AppCleanUpStepTest`. The new test cases cover apps with and without the queue label, the `SchedulingFailure` spec-build failure path, and the `resourceRetainPolicy` path.

Generated-by: Claude Opus 5


spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppCleanUpStep.java
```diff
@@ -163,6 +165,10 @@ public ReconcileProgress reconcile(
     for (HasMetadata resource : resourcesToRemove) {
       ReconcilerUtils.deleteResourceIfExists(context.getClient(), resource, forceDelete);
     }
+    if (KueueWorkloadFactory.hasQueueName(application)) {
```
peter-toth: **Finding 1.** There is one path between the cleanup decision and this line that returns without reaching it. `isReleasingResourcesForSchedulingFailureAttempt` re-computes the spec, and when that throws:

```java
      } catch (Exception e) {
        ...
        ApplicationState updatedState =
            new ApplicationState(
                ApplicationStateSummary.ResourceReleased,
                "Cannot build Spark spec for given application, "
                    + "consider all resources as released.");
        ...
        return appendStateAndRequeueAfter(   // :157, above this line
```

That is not a dead branch. It fires only when the last observed state before termination was `SchedulingFailure` - i.e. precisely when the spec was already rejected once - so a re-build failing the second time is the expected case, not the exotic one.

And it is terminal. The state it writes is `ResourceReleased`, so on every later reconcile `checkEarlyExitForTerminatedApp` takes its `:237-241` branch and returns before the release can run. `ttlAfterStopMillis` defaults to `-1L`, so the CR is never garbage collected either, and the `ownerReference` never fires. The `Workload` stays `Admitted` and holds the `ClusterQueue` quota until someone deletes the app by hand - the exact scenario the "Why" section says this PR prevents.

I confirmed it with a probe test: a labelled app whose last state is `SchedulingFailure` and whose `getDriverPreResourcesSpec()` throws, asserting `KueueWorkloadUtils` is never touched. It passes on this head.

Hoisting the block above `resourcesToRemove` fixes it, and is safe because the release does not depend on which resources get deleted:

```java
    if (KueueWorkloadFactory.hasQueueName(application)) {
      // Release the quota. A restarted attempt is queued again with a new Workload.
      KueueWorkloadUtils.releaseWorkload(context.getClient(), application);
    }
    List<HasMetadata> resourcesToRemove = new ArrayList<>();
```

I applied exactly that and re-ran `AppCleanUpStepTest`: the probe flips to failing, which is the point, and all 17 pre-existing tests stay green - including the three `verify(mockApp).getMetadata()` assertions you added, since the call count is unchanged. It also closes the same hole for a `deleteResourceIfExists` that throws, and it makes the comment you added to those three tests literally true, since the label lookup would then really happen before the resources are released.

Returning the quota a moment before the driver pod finishes terminating is fine: `ClusterQueue` quota is Kueue's bookkeeping, not a kube-scheduler reservation, and Kueue's own integrations release on job completion while pods drain asynchronously.

spark-operator/src/test/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppCleanUpStepTest.java
```diff
@@ -115,6 +118,50 @@ class AppCleanUpStepTest {
...
+    app.setStatus(prepareApplicationStatus(ApplicationStateSummary.RunningHealthy));
+    SparkAppContext mockAppContext = mock(SparkAppContext.class);
+    when(mockAppContext.getResource()).thenReturn(app);
+    when(mockAppContext.getClient()).thenReturn(mock(KubernetesClient.class));
+    when(mockAppContext.getDriverPod()).thenReturn(Optional.empty());
+    when(mockRecorder.appendNewStateAndPersist(eq(mockAppContext), any())).thenReturn(true);
+
+    try (MockedStatic<KueueWorkloadUtils> kueue = Mockito.mockStatic(KueueWorkloadUtils.class)) {
+      cleanUpWithReason.reconcile(mockAppContext, mockRecorder);
+      kueue.verifyNoInteractions();
```
peter-toth, line 161: **Finding 2.** These two tests pin the label check in both directions, which is the right pair. What is not pinned is the other half of the contract the description states:

> Apps that keep their resources via `resourceRetainPolicy` keep their quota until the resources are released.

Today that holds structurally - `:129` and `:281` both return above the release - so there is nothing to catch. It becomes worth a test the moment finding 1 is applied, because that change moves the release upwards and the only thing keeping the retain guarantee is that it stays below `:113-132`. A third sibling of these two would guard it:

```java
  @Test
  void cleanupWithRetainPolicyKeepsKueueWorkload() {
    SparkAppStatusRecorder mockRecorder = mock(SparkAppStatusRecorder.class);
    AppCleanUpStep cleanUpStep = new AppCleanUpStep();   // not on demand
    SparkApplication app = new SparkApplication();
    app.setMetadata(
        new ObjectMetaBuilder()
            .withName("app1")
            .withNamespace("default")
            .withLabels(Map.of(Constants.LABEL_QUEUE_NAME, "test-queue"))
            .build());
    app.getSpec().getApplicationTolerations().setResourceRetainPolicy(ResourceRetainPolicy.Always);
    app.getSpec().getApplicationTolerations().setRestartConfig(
        new RestartConfig(RestartPolicy.Never, 0, 0L, 0L));
    app.setStatus(prepareApplicationStatus(ApplicationStateSummary.Failed));
    ...
    try (MockedStatic<KueueWorkloadUtils> kueue = Mockito.mockStatic(KueueWorkloadUtils.class)) {
      cleanUpStep.reconcile(mockAppContext, mockRecorder);
      kueue.verifyNoInteractions();
    }
  }
```

`Failed` is `isStopping()` and `Always` retains, so with `RestartPolicy.Never` the step should take the `TerminatedWithoutReleaseResources` branch and never touch the `Workload`. Adjust the tolerations wiring to whatever the setters look like - the shape is what matters.

peter-toth (commented):

I walked every path into the release and the placement is right on all but one: the terminated, on-demand, stopping-with-release and stopping-with-restart paths all reach it, while the two retain paths early-return above it, which is exactly what the description promises. The exception is the `SchedulingFailure` re-compute branch - when the spec can no longer be built it returns from inside the `catch`, above the release, and sets `ResourceReleased`, after which `checkEarlyExitForTerminatedApp` never lets the release run again. With `ttlAfterStopMillis = -1` the CR is never collected either, so that is the indefinite quota hold this PR exists to prevent.

### Blocking
- **1.** *The `SchedulingFailure` spec-build failure returns above the release:* `AppCleanUpStep.java:150-158` returns `ResourceReleased` from the `catch`, so the `Workload` is never deleted and no later reconcile reaches line 168. Hoisting the release above `resourcesToRemove` closes it; I applied that and confirmed it. [inline: [`AppCleanUpStep.java:168`](https://github.com/apache/spark-kubernetes-operator/pull/837#discussion_r4034542305)]

### Non-blocking
- **2.** *Nothing pins the retain path keeping its quota:* the description promises it and today an early return guarantees it, but the fix in finding 1 moves the release upwards, so the guarantee becomes worth a test rather than an accident of ordering. [inline: [`AppCleanUpStepTest.java:161`](https://github.com/apache/spark-kubernetes-operator/pull/837#discussion_r4034542309)]

spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppCleanUpStep.java
dongjoon-hyun: Thank you for the careful analysis, @peter-toth. You're right. I moved the release above `resourcesToRemove` and added `cleanupReleasesKueueWorkloadWhenSpecBuildFails` as a regression test. It fails before this change and passes after it.

spark-operator/src/test/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppCleanUpStepTest.java
dongjoon-hyun: Agreed. I added `cleanupWithRetainPolicyKeepsKueueWorkload` to pin that the retain path keeps the `Workload` and ends in `TerminatedWithoutReleaseResources`.

peter-toth (approved):

dongjoon-hyun:
Thank you, @peter-toth !

dongjoon-hyun:
Merged to main

PR #838, opened 2026-09-16
[SPARK-59537] Honor `operatorRbac.configManagement.create` in Helm chart

dongjoon-hyun:

This PR aims to honor `operatorRbac.configManagement.create` in the Helm chart.

Previously, this value was ignored, so setting it to `false` did not prevent the config-monitor `Role` and `RoleBinding` from being created.

Yes, `operatorRbac.configManagement.create=false` now takes effect. The default is `true`, so the default behavior is unchanged.

Pass the CIs and manual `helm template` checks.

Generated-by: Claude Opus 5


peter-toth (approved):

dongjoon-hyun:
Thank you, @peter-toth !

dongjoon-hyun:
Merged to main

PR #840, opened 2026-09-17
[SPARK-59606] Map `SparkCluster` pods by `spark.operator/spark-cluster-name` in `SparkClusterReconciler`

dongjoon-hyun:

This PR aims to map `SparkCluster` pods by the `spark.operator/spark-cluster-name` label instead of `spark.operator/spark-app-name` in `SparkClusterReconciler`.

`SparkCluster` master and worker pods don't have the `spark.operator/spark-app-name` label, so their events never reached the owning `SparkCluster`.

Yes. Master and worker pod events now trigger `SparkCluster` reconciliation.

Pass the CIs with the newly added test case.

Generated-by: Claude Opus 5


dongjoon-hyun:
Thank you, @peter-toth . I created this PR based on your finding and JIRA issue.
- https://github.com/apache/spark-kubernetes-operator/pull/836

dongjoon-hyun:
Let me merge this to unblock other PRs.

dongjoon-hyun:
Merged to main

spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/SparkClusterReconciler.java
```diff
@@ -155,7 +154,7 @@ public List<EventSource<?, SparkCluster>> prepareEventSources(
         new InformerEventSource<>(
             InformerEventSourceConfiguration.from(Pod.class, SparkCluster.class)
                 .withSecondaryToPrimaryMapper(
-                    basicLabelSecondaryToPrimaryMapper(LABEL_SPARK_APPLICATION_NAME))
+                    basicLabelSecondaryToPrimaryMapper(LABEL_SPARK_CLUSTER_NAME))
                 .withLabelSelector(commonResourceLabelsStr())
```
peter-toth, line 158: **Finding 1.** The mapper now reads the right label, but no cluster pod ever reaches it. This selector expands to `spark.operator/name=spark-kubernetes-operator`, and master and worker pods do not carry that label.

`SparkClusterResourceSpecFactory` hands `sparkClusterResourceLabels(cluster)` to `OwnerResourceDecorator`, which writes them to the StatefulSet's *own* `metadata.labels`. The pod template is untouched. `SparkClusterResourceSpec` gives it only `spark-role`, `spark.operator/spark-cluster-name` and `spark-version` (`spark-submission-worker/src/main/java/org/apache/spark/k8s/operator/SparkClusterResourceSpec.java:233-237` and `:315-319`), and Kubernetes does not copy StatefulSet labels onto its pods. JOSDK passes the configured selector to `FilterWatchListDeletable.withLabelSelector` before creating the informer, so these pods are never listed or watched.

The app side works only because it takes a different path: `SparkAppResourceSpecFactory.overrideDependencyConf` pushes every `sparkAppResourceLabels` entry through `spark.kubernetes.driver.label.*` and `spark.kubernetes.executor.label.*`, so driver and executor pods really do carry `spark.operator/name`.

What I ran at `c740424`: built the real spec through `SparkClusterResourceSpecFactory.buildResourceSpec` and compared the master pod-template labels against the selector this method configures.

    ### selector          = spark.operator/name=spark-kubernetes-operator
    ### master pod labels = {spark-role=master, spark.operator/spark-cluster-name=cluster-1}

The smallest fix, which I verified makes the selector match those labels:

```suggestion
                .withLabelSelector(LABEL_SPARK_CLUSTER_NAME)
```

The static import of `commonResourceLabelsStr` has to go with it. Left in place it is unused, and `spotlessCheck` fails on it.

The alternative is to add `commonManagedResourceLabels()` to the two pod templates and keep this selector. That keeps both pod informers on one shared watch, since JOSDK's `InformerClassifier` keys the informer pool on the label selector and diverging selectors mean a second Pod watch. It also keeps the informer scoped to this operator instance. The cost is a change in `spark-submission-worker`. Existing `RunningHealthy` clusters would not roll, because `getReconcileSteps` re-applies the StatefulSets only from `Submitted`.

peter-toth (commented):

The mapper swap is correct, and the new test really does fail on the unfixed code. I ran it at `6f2b3c7`: 12 tests, only `podInformerMapsPodsBySparkClusterNameLabel` fails. But the mapper is not the only thing standing between a cluster pod and a `SparkCluster` reconcile. One line below it, the informer's own label selector is `commonResourceLabelsStr()`, which expands to `spark.operator/name=spark-kubernetes-operator`. Master and worker pods never carry that label, so no cluster pod reaches the mapper and the description's user-facing claim does not hold yet.

That correction is mostly on me. SPARK-59606 claims "the pods still match the informer's `commonResourceLabelsStr()` selector and are cached". I checked it today and that sentence is wrong.

This landed as `3a2f4f5` before I finished, so finding 1 needs a follow-up.

### Blocking
- **1.** *Pod informer's selector filters out every cluster pod:* `OwnerResourceDecorator` writes the common labels to the StatefulSet's own metadata, not to its pod template, so cluster pods carry only `spark-role`, `spark.operator/spark-cluster-name` and `spark-version`. Switching the selector to the existence form the `Workload` informer three lines below already uses is the fix I verified. [inline: `SparkClusterReconciler.java:158`](https://github.com/apache/spark-kubernetes-operator/pull/840#discussion_r4036172806)

### Non-blocking
- **2.** *The new test pins the mapper but leaves the selector unpinned:* the sibling `kueueWorkloadInformerIsRegisteredOnlyWhenEnabled` asserts its own selector, and that is the assertion that would have caught finding 1. [inline: `SparkClusterReconcilerTest.java:299`](https://github.com/apache/spark-kubernetes-operator/pull/840#discussion_r4036172817)
- **3.** *"Master and worker pod events now trigger `SparkCluster` reconciliation" promises more than it delivers:* nothing derives cluster status from pods. `getReconcileSteps` adds no step for `RunningHealthy` (`SparkClusterReconciler.java:186-193`), there is no cluster counterpart to the `AppDriver*Observer` family, and `ReconcilerUtils.toUpdateControl` returns `noUpdate().rescheduleAfter(...)`. So once the selector is fixed, a pod event on a healthy cluster only resets the 120s requeue timer. The real gain is waking a `Submitted` cluster's `ClusterInitStep` retry, which is worth saying instead.

spark-operator/src/test/java/org/apache/spark/k8s/operator/reconciler/SparkClusterReconcilerTest.java
```diff
@@ -274,6 +274,32 @@ private EventRecord captureRecordedEvent() {
...
+      Pod pod = new Pod();
+      pod.setMetadata(
+          new ObjectMetaBuilder()
+              .withName("cluster-1-master-0")
+              .withNamespace("default")
+              .withLabels(Map.of(LABEL_SPARK_CLUSTER_NAME, "cluster-1"))
+              .build());
+      assertEquals(
+          Set.of(new ResourceID("cluster-1", "default")),
+          podConfig.getSecondaryToPrimaryMapper().toPrimaryResourceIDs(pod));
```
peter-toth, line 299: **Finding 2.** This pins the mapper but not the selector, and the selector is where finding 1 lives. `kueueWorkloadInformerIsRegisteredOnlyWhenEnabled` already asserts its own selector, so the pod source is the odd one out.

Adding this next to the `podConfig` cast fails at this head with `expected: <spark.operator/spark-cluster-name> but was: <spark.operator/name=spark-kubernetes-operator>`, and passes once finding 1 is fixed:

```java
      assertEquals(LABEL_SPARK_CLUSTER_NAME, podConfig.getInformerConfig().getLabelSelector());
```

A stronger version, and one that survives either fix for finding 1, builds the real pod-template labels through `SparkClusterResourceSpecFactory.buildResourceSpec` and asserts the selector is satisfied by them. That breaks if either side drifts, which is the invariant that actually matters here.

dongjoon-hyun:
Oh, thank you, @peter-toth . I'll make a follow-up PR.

peter-toth:
Sorry, @dongjoon-hyun , my review took quite some time, could you please check the findings if you agree with them?

dongjoon-hyun:
Here is the follow-up PR.
- https://github.com/apache/spark-kubernetes-operator/pull/842

PR #841, opened 2026-09-17
[SPARK-59610] Fill missing `master` and `worker` requests of Kueue `Workload` for `SparkCluster`

dongjoon-hyun:

This PR aims to fill in the missing CPU and memory requests of the `master` and `worker` `PodSet`s when `KueueWorkloadFactory` builds a Kueue `Workload` for a `SparkCluster`.

A missing request is filled in as follows. Requests that are already in the pod template are kept, and a missing request defaults to the limit, like Kubernetes.

| PodSet | CPU | Memory |
|---|---|---|
| `master` | `1` | `SPARK_DAEMON_MEMORY` (Spark default `1g`) + overhead |
| `worker` | `SPARK_WORKER_CORES`, or `1` if unset | `SPARK_DAEMON_MEMORY` (Spark default `1g`) + overhead + `SPARK_WORKER_MEMORY` (if set) |

- No new configuration is introduced. The values come from the same environment variables that Spark standalone reads (`SparkClassCommandBuilder` and `WorkerArguments`).
- Only literal `value`s of the container `env` are used. Anything the operator cannot read at build time is ignored: `valueFrom`, `envFrom`, and `conf/spark-env.sh` inside the image.
- Like Spark, a memory string without a unit is in bytes.
- The overhead is `max(0.10 * SPARK_DAEMON_MEMORY, 384MiB)`, the default rule of the `SparkApplication` pod sets. It is applied to the daemon heap only because `SPARK_WORKER_MEMORY` is shared by an unknown number of executor processes.
- Spark standalone has no setting for the number of master cores, so `1` is used unless the pod template specifies it.
- The worker `1` CPU and the omission of `SPARK_WORKER_MEMORY` are minimum values chosen by the operator, not Spark defaults. When unset, Spark standalone infers the worker cores from `availableProcessors()` and the worker memory from the total memory minus 1 GiB, which the operator cannot know. So the operator logs a warning when a worker has neither `SPARK_WORKER_CORES` and `SPARK_WORKER_MEMORY` nor CPU and memory limits.

Unlike the driver and executor pods of `SparkApplication`, Spark does not set any requests on the master and worker pods. So far, `KueueWorkloadFactory` copied the `StatefulSet` pod templates into the `PodSet`s as they were. As a result, a `SparkCluster` without `resources.requests` in its templates produced a `Workload` with zero usage, and Kueue admitted it without accounting it against the `ClusterQueue` quota.

No. The Kueue integration is not released yet.

Pass the CIs with the newly added and updated test cases in `KueueWorkloadFactoryTest`.

- `testBuildWorkloadForSparkCluster` (updated)
- `testBuildWorkloadForSparkClusterFillsMissingRequests`
- `testBuildWorkloadForSparkClusterUsesLimitsAsMissingRequests`
- `testCalculateDaemonMemoryMiB`

Generated-by: Claude Opus 5


dongjoon-hyun:
All tests passed.

dongjoon-hyun:
Could you review this PR too, @peter-toth ?

peter-toth:
Checking it now. My skill is a bit slow sorry.

dongjoon-hyun:
Oh, no problem at all. Sorry for asking too many reviews, @peter-toth .

spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadFactory.java
```diff
@@ -188,14 +189,46 @@ private static Workload buildWorkload(
...
+   */
+  private static PodSet buildPodSet(final String role, final StatefulSet statefulSet) {
+    PodTemplateSpec templateSpec = statefulSet.getSpec().getTemplate();
+    // The operator always creates the container named after the role.
+    Container container = selectContainer(templateSpec.getSpec(), role, role);
+    Map<String, Quantity> requests = getOrCreateRequests(container);
+    Map<String, Quantity> limits = container.getResources().getLimits();
+    Map<String, String> env = getEnv(container);
+    boolean isWorker = PODSET_WORKER.equals(role);
+    String cpu = isWorker ? env.getOrDefault("SPARK_WORKER_CORES", DEFAULT_CORES) : DEFAULT_CORES;
```
peter-toth, line 207: **Finding 1.** Spark standalone does not default the worker to one core. `WorkerArguments` sets `var cores = inferDefaultCores()` and only overrides it when the env var is present:

```scala
  def inferDefaultCores(): Int = {
    Runtime.getRuntime.availableProcessors()
  }

  def inferDefaultMemory(): Int = {
    ...
    // Leave out 1 GB for the operating system, but don't return a negative memory size
    math.max(totalMb - 1024, Utils.DEFAULT_DRIVER_MEM_MB)   // DEFAULT_DRIVER_MEM_MB == 1024
  }
```

`SPARK_WORKER_MEMORY` is the same story. Unset does not mean zero; it means the node's total memory minus 1 GiB. And with no CPU or memory limit on the container there is no cgroup bound for the JVM to read, so both inferences see the node rather than the pod.

Here is what a plain `SparkCluster` produces at this head, with no `resources` and no env in either template, i.e. the input the description names as the bug:

```
### podSet=master count=1 container=master requests={cpu=1, memory=1408Mi} limits={}
### podSet=worker count=2 container=worker requests={cpu=1, memory=1408Mi} limits={}
```

The master numbers are right: its heap really is `SPARK_DAEMON_MEMORY` and it reserves no cores. The worker numbers are a floor. A `ClusterQueue` with a 16-CPU quota would admit 16 of these workers, each of which advertises the whole node to executors, so the quota is still not enforced for the case the PR targets.

Two ways out that fit in this PR:

1. State the limitation. Correct the description's table, and `log.warn` here when the worker container has neither `SPARK_WORKER_CORES` / `SPARK_WORKER_MEMORY` nor a CPU / memory limit, so whoever reads the operator log knows the Workload is a floor.
2. Refuse the shape, which is what this file already does for the other two things it cannot account for. `buildWorkload` throws `UnsupportedOperationException` for a `HorizontalPodAutoscaler` and for dynamic allocation, both because a count is not knowable. A worker whose size is not knowable is the same class of problem. Harsher, since it rejects the default cluster outright.

Finding 4 is the third way: make the size knowable.

On what I checked how: the probe output above I ran at `84df491`. That a worker pod with no limits really sees the node's CPUs and memory is read off `WorkerArguments` plus the JVM's container support, not observed on a cluster. Deploying a default `SparkCluster` on a multi-core node and reading the `Cores` figure on the worker's web UI would settle it.

peter-toth (commented):

Filling the requests explicitly is the right call. Kueue reads the PodSet template without the API server's limits-to-requests defaulting, so a template carrying only limits would still have counted as zero. The byte-without-a-unit rule matches Spark exactly: `Utils.memoryStringToMb` is `byteStringAsBytes(str) / 1024 / 1024`. All three cluster tests fail on `6920cc5`, so they pin the change.

My one objection is the worker default. Spark standalone does not default `SPARK_WORKER_CORES` to `1`; it defaults to `Runtime.getRuntime.availableProcessors()`, and `SPARK_WORKER_MEMORY` defaults to the node's memory minus 1 GiB. So the plainest `SparkCluster`, the case this PR exists to fix, still produces a Workload that accounts almost nothing.

### Blocking
- **1.** *Worker accounting understates the default configuration, and the description's Spark defaults are wrong:* `WorkerArguments.inferDefaultCores()` is `availableProcessors()` and `inferDefaultMemory()` is `max(totalMb - 1024, 1024)`. A default cluster's worker PodSet asks for `cpu: 1, memory: 1408Mi` with no limits set, so each worker will advertise the whole node. [inline: `KueueWorkloadFactory.java:207`](https://github.com/apache/spark-kubernetes-operator/pull/841#discussion_r4036382704)

### Non-blocking
- **2.** *The overhead treats the executor pool as one JVM:* `0.10 * (daemon + workerMemory)` adds 921Mi for an 8g pool, where eight separate executor processes would each want their own 384Mi floor. [inline: `KueueWorkloadFactory.java:350`](https://github.com/apache/spark-kubernetes-operator/pull/841#discussion_r4036382711)
- **3.** *`getEnv`'s contract is narrower than its javadoc says:* `envFrom` is ignored too, and so is `conf/spark-env.sh`, which is where Spark's own template documents these three variables. [inline: `KueueWorkloadFactory.java:481`](https://github.com/apache/spark-kubernetes-operator/pull/841#discussion_r4036382719)

### Alternatives
- **4.** *Write the worker's capacity instead of guessing it:* if `SparkClusterResourceSpec` set `SPARK_WORKER_CORES` and `SPARK_WORKER_MEMORY` from the worker container's CPU and memory limit, Spark's advertised capacity and Kueue's accounting would agree by construction, and `buildPodSet` would need no env reading at all. [inline: `KueueWorkloadFactory.java:199`](https://github.com/apache/spark-kubernetes-operator/pull/841#discussion_r4036382735)

spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadFactory.java
```diff
@@ -300,6 +333,23 @@ static long calculateDriverMemoryMiB(
...
+   */
+  static long calculateDaemonMemoryMiB(final String daemonMemory, final String workerMemory) {
+    // Like Spark, the memory is in bytes unless otherwise specified.
+    String memory = StringUtils.isEmpty(daemonMemory) ? DEFAULT_MEMORY : daemonMemory;
+    long memMiB = JavaUtils.byteStringAsBytes(memory) / 1024 / 1024;
+    if (StringUtils.isNotEmpty(workerMemory)) {
+      memMiB += JavaUtils.byteStringAsBytes(workerMemory) / 1024 / 1024;
+    }
+    long minOverheadMiB = JavaUtils.byteStringAsMb(DEFAULT_MIN_MEMORY_OVERHEAD);
+    return memMiB + Math.max((long) (DEFAULT_MEMORY_OVERHEAD_FACTOR * memMiB), minOverheadMiB);
```
peter-toth: **Finding 2.** The overhead rule is borrowed from the `SparkApplication` pod sets, where `memMiB` is one JVM's heap. Here `memMiB` is `daemonHeap + SPARK_WORKER_MEMORY`, and `SPARK_WORKER_MEMORY` is a pool the worker hands out to N separate executor processes, each of which needs its own non-heap headroom.

With `SPARK_WORKER_MEMORY=8g` split into eight 1g executors this returns `1024 + 8192 + 921`. Spark's own Kubernetes rule for those same eight executors would add `8 * max(0.1 * 1024, 384) = 3072Mi`, so the pod is short about 2.1 GiB of the headroom Spark would have asked for.

The executor count is not knowable here, so I am not asking for an exact number. Applying the factor to the daemon heap only and leaving `SPARK_WORKER_MEMORY` to carry its own executors would at least not imply a precision the formula does not have:

```java
    long memMiB = JavaUtils.byteStringAsBytes(memory) / 1024 / 1024;
    long minOverheadMiB = JavaUtils.byteStringAsMb(DEFAULT_MIN_MEMORY_OVERHEAD);
    long total = memMiB + Math.max((long) (DEFAULT_MEMORY_OVERHEAD_FACTOR * memMiB), minOverheadMiB);
    if (StringUtils.isNotEmpty(workerMemory)) {
      total += JavaUtils.byteStringAsBytes(workerMemory) / 1024 / 1024;
    }
    return total;
```

That is a smaller number than what is there now for a large pool, so if you prefer the current shape a comment saying the factor is a pool-wide approximation would do.

spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadFactory.java
```diff
@@ -437,6 +478,31 @@ private static void decorateContainerResources(
     }
   }
 
+  /** Returns the environment variables with the literal values. `valueFrom` is ignored. */
```
peter-toth: **Finding 3.** `valueFrom` is not the only source this misses. `envFrom` (a `ConfigMap` or `Secret` env source on the container) is also invisible, and so is `conf/spark-env.sh`, which is the place Spark's own template documents for exactly these three variables:

```
# - SPARK_WORKER_CORES, to set the number of cores to use on this machine
# - SPARK_WORKER_MEMORY, to set how much total memory workers have to give executors (e.g. 1000m, 2g)
# - SPARK_DAEMON_MEMORY, to allocate to the master, worker and history server themselves (default: 1g).
```

Both `sbin/start-worker.sh` and `sbin/spark-daemon.sh` source `bin/load-spark-env.sh`, so a `spark-env.sh` baked into a custom image is read by the worker and cannot be read by the operator. That is not fixable here, but the comment should say what the accounting actually covers rather than naming one of three gaps:

```suggestion
  /**
   * Returns the environment variables with the literal `value`s of the container. Anything the
   * operator cannot read at build time is ignored: `valueFrom`, `envFrom`, and `conf/spark-env.sh`
   * inside the image, which `load-spark-env.sh` sources for the master and worker daemons.
   */
```

The description's matching bullet says only "Only literal `value`s of the container `env` are used. `valueFrom` is ignored" and would want the same widening.

spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadFactory.java
```diff
@@ -188,14 +189,46 @@ private static Workload buildWorkload(
...
 
-  private static PodSet buildPodSet(final String name, final StatefulSet statefulSet) {
+  /**
+   * Builds a PodSet for a Spark standalone role (`master` or `worker`) from the StatefulSet. Unlike
+   * the SparkApplication pods, Spark does not set the requests of the master and worker pods, so
+   * the missing CPU and memory requests are calculated from the environment variables of the
+   * container in the same way as Spark standalone does. Otherwise, Kueue admits the pods without
+   * accounting them against the quota.
+   */
+  private static PodSet buildPodSet(final String role, final StatefulSet statefulSet) {
```
peter-toth, line 199: **Finding 4.** An alternative worth considering, as a follow-up rather than in this PR: instead of reading the env to predict what Spark will infer, have the operator *write* it.

`SparkClusterResourceSpec.buildWorkerStatefulSet` already owns the worker container. If it set the two variables from the container's own resources whenever they are absent, roughly:

```java
    // in buildWorkerStatefulSet, from the worker container's limits, falling back to requests
    Quantity cpu = limitOrRequest(container, "cpu");
    Quantity mem = limitOrRequest(container, "memory");
    if (cpu != null) addEnvIfAbsent("SPARK_WORKER_CORES", cores(cpu));
    if (mem != null) addEnvIfAbsent("SPARK_WORKER_MEMORY", (mib(mem) - daemonMiB - overheadMiB) + "m");
```

then three things follow. The worker's advertised capacity stops depending on which node it lands on, which is worth having on its own. Spark's view and Kueue's accounting agree by construction, so finding 1 goes away. And `buildPodSet` no longer needs `getEnv` at all: the requests are always present, and it reduces to the `fillMissingRequest` limits fallback.

The honest counter-arguments:

- It changes the applied `StatefulSet`, so an existing cluster's worker capacity would go from node-sized to limit-sized on upgrade. That is arguably a fix, but it is a behavior change and belongs in its own ticket.
- It spans `spark-submission-worker`, well outside this PR.
- A worker with no limits at all still has nothing to derive from, so option 1 or 2 from finding 1 is still needed for that case.
- Deliberate oversubscription stays possible only if the operator fills the variables when absent rather than overwriting them, which is why the sketch guards on absence.

dongjoon-hyun:
Thank you for the detailed review, @peter-toth . I addressed the comments.

1. Worker default: I took the first option. The operator now logs a warning when a worker has neither `SPARK_WORKER_CORES` and `SPARK_WORKER_MEMORY` nor CPU and memory limits. I also fixed the PR description so that the worker `1` CPU is described as a minimum chosen by the operator, not as a Spark default.
2. Overhead: I applied your suggestion. The overhead is now applied to the daemon heap only, and `SPARK_WORKER_MEMORY` is added as it is.
3. `getEnv`: I applied your Javadoc suggestion and widened the matching bullet in the PR description.
4. Writing `SPARK_WORKER_CORES` and `SPARK_WORKER_MEMORY` from the worker container resources is a good idea. Since it changes the applied `StatefulSet` and touches `spark-submission-worker`, let me handle it in a separate JIRA issue as a follow-up. [SPARK-59611 Set `SPARK_WORKER_CORES` and `SPARK_WORKER_MEMORY` if absent from `worker` container resources in `SparkCluster`](https://issues.apache.org/jira/browse/SPARK-59611)

peter-toth (approved):

dongjoon-hyun:
Thank you always for your support, @peter-toth !

dongjoon-hyun:
Merged to main

PR #842, opened 2026-09-17
[SPARK-59606][FOLLOWUP] Use `spark.operator/spark-cluster-name` label selector for `SparkCluster` pod informer

dongjoon-hyun:

This PR is a follow-up of #840 to change the label selector of the `Pod` informer in `SparkClusterReconciler` from `commonResourceLabelsStr()` to the existence selector `spark.operator/spark-cluster-name`, like the Kueue `Workload` informer.

In addition, this PR adds a test assertion on the `Pod` informer's label selector.

#840 fixed the secondary-to-primary mapper, but no `SparkCluster` pod reaches the mapper yet. The informer selector `commonResourceLabelsStr()` expands to `spark.operator/name=spark-kubernetes-operator`. However, `OwnerResourceDecorator` writes these common labels only to the `StatefulSet` metadata, not to the pod templates. So master and worker pods only have the `spark-role`, `spark.operator/spark-cluster-name`, and `spark-version` labels, and the informer never lists or watches them.

This PR doesn't add the common labels to the master and worker pod templates instead, because that change would touch `spark-submission-worker` and wouldn't reach the pods of existing `RunningHealthy` clusters, whose `StatefulSet`s are not re-applied.

Yes. The `Pod` informer now watches `SparkCluster` master and worker pods. Their events now requeue the owning `SparkCluster`, so a `Submitted` cluster retries `ClusterInitStep` immediately instead of waiting for the next scheduled requeue.

Pass the CIs with the updated test case. The new assertion fails without the fix:

```
expected: <spark.operator/spark-cluster-name> but was: <spark.operator/name=spark-kubernetes-operator>
```

Generated-by: Claude Opus 5


peter-toth (approved):

peter-toth:
Thank you for the follow-up!

dongjoon-hyun:
Thank you, @peter-toth !

dongjoon-hyun:
Merged to main

PR #843, opened 2026-09-17
[SPARK-59612] Wait for Kueue `Workload` admission in `AppInitStep` and `ClusterInitStep`

dongjoon-hyun:

This PR aims to make `AppInitStep` and `ClusterInitStep` wait for the Kueue `Workload` admission of a `SparkApplication` or `SparkCluster` labeled with `kueue.x-k8s.io/queue-name`.

| Resource | State | With `kueue.x-k8s.io/queue-name` |
|---|---|---|
| `SparkApplication` | `Submitted`, `ScheduledToRestart` | Hold driver creation until the `Workload` is admitted |
| `SparkCluster` | `Submitted` | Hold master / worker creation until the `Workload` is admitted |

- The check runs right after the `spec.suspend` hold and the restart backoff, and uses `KueueWorkloadUtils.requestAdmission` (SPARK-59588).
  - `ADMITTED`: create the resources as before.
  - `PENDING`: requeue with the default interval. The `Workload` informer (SPARK-59593) triggers a reconciliation on admission.
  - `STALE`: requeue after 5 seconds.
- A Kueue API failure of the admission request, such as missing RBAC or an unreachable Kueue, is retried after 5 seconds. It does not become `SchedulingFailure`, which is terminal for a `SparkCluster`. An unsupported spec, such as dynamic allocation, still ends up in `SchedulingFailure`.
- `spec.suspend` takes precedence. A suspended resource does not get a `Workload`, and suspending a queued resource deletes its `Workload` so that it does not keep holding the quota.
- Like the `spec.suspend` hold, the check is skipped once the driver pod (or the master `StatefulSet`) already exists.
- No state is written while waiting, so the resource stays in `Submitted` / `ScheduledToRestart`.
- Filter the `Workload` informer events down to admission changes and deletions. The operator creates the `Workload` itself, and the other status updates of a pending `Workload` would use up the per-resource reconciliation rate limit (5 per 15 seconds).
- Add a `Kueue` section to `docs/spark_custom_resources.md`, including the current limitations:
  - A `SparkCluster` without CPU or memory settings on its `worker` container requests only the default, while the worker advertises the whole node to its executors.
  - Preemption is not honored yet. A later eviction releases the quota while the pods keep running.
  - A queued resource spends one more reconciliation on the admission itself. So a `SparkApplication` that finishes within the first rate limit window can report its terminal state late and record `Failed` before `Succeeded`.
- Extend `tests/e2e/kueue` with labeled `SparkApplication` and `SparkCluster`. Each waits on a zero-quota `ClusterQueue` and completes after the quota is raised. The labeled `SparkApplication` runs `SparkPi` with more slices to outlast the first rate limit window. That limitation is not specific to Kueue and will be addressed separately.

To wire the existing Kueue building blocks (SPARK-59588, SPARK-59593, SPARK-59594, SPARK-59610) into reconciliation, so that the operator actually participates in Kueue job queueing.

No. The Kueue integration is not released yet.

Pass the CIs with the newly added test cases in `AppInitStepTest`, `ClusterInitStepTest`, `KueueWorkloadUtilsTest`, `SparkAppReconcilerTest` and `SparkClusterReconcilerTest`, and the extended `tests/e2e/kueue` suite.

Generated-by: Claude Opus 5


spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppInitStep.java
```diff
@@ -100,6 +103,20 @@ public ReconcileProgress reconcile(
       }
     }
     try {
+      // Like the suspend hold, a driver requested before must not be left unobserved.
+      if (KueueWorkloadFactory.hasQueueName(app) && !isDriverRequested(context)) {
```
peter-toth: **Finding 1.** This block is below the suspend gate at `:73`, and `buildWorkload` has no other caller in `src/main`. Two consequences.

First, `KueueWorkloadFactory.java:186` computes `.active(!suspend)`, but `suspend` is always `false` by the time it runs, so nothing ever writes `active: false`. That wiring is unreachable today.

Second, and the reason this matters: suspending a resource that is *already* queued does not release its quota. `getOrCreateSecondaryResource` only creates, never updates, and `releaseWorkload` is called from `AppCleanUpStep` alone, which a `Submitted` resource never reaches. So the `Workload` sits there with `active: true`, Kueue admits it, and the quota is consumed by an application the operator is deliberately refusing to start.

Observed at `a038938` with the mock client, reconciling once, then setting `suspend: true` and reconciling again:

```
### r1 (not suspended) -> ReconcileProgress(completed=true, requeue=true, requeueAfterDuration=PT2M)
### after r1: workload=true active=true
### r2 (suspended)     -> ReconcileProgress(completed=true, requeue=true, requeueAfterDuration=PT2M)
### after r2: workload=true active=true
```

It also makes `docs/spark_custom_resources.md:602` ("A suspended resource does not get a `Workload` at all") true only when `suspend` was set before the first admission request.

The smallest fix, which I applied and re-ran:

```java
    if (app.getSpec().isSuspend() && !isDriverRequested(context)) {
      log.debug("Application is suspended, driver resources would not be requested.");
      if (KueueWorkloadFactory.hasQueueName(app)) {
        // A resource suspended while queued must not keep holding the Kueue quota.
        KueueWorkloadUtils.releaseWorkload(context.getClient(), app);
      }
      return completeAndDefaultRequeue();
    }
```

`after r2` then reads `workload=false`. Note this needs one change in your own new test: `suspendedAppWithQueueNameDoesNotCreateKueueWorkload` asserts `verify(mockContext, never()).getClient()` and never stubs it, so it NPEs at `KueueWorkloadUtils:142`. Stub the client and drop that `never()` line; the `assertNull(getWorkload())` it already has is the assertion that matters. `ClusterInitStep.java:68` needs the same treatment.

The better shape, if you want `active` to earn its keep, is to move the admission block above the suspend gate and have `requestAdmission` reconcile `spec.active` on an existing `Workload`. That keeps the queue position across a suspend and is what Kueue's `active` field is for. It is more work, and deleting is enough to stop the leak.

spark-operator/src/test/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppInitStepTest.java
```diff
@@ -598,4 +612,182 @@ void staleInformerSnapshotDoesNotBypassSuspend() {
...
+
+    ReconcileProgress progress = appInitStep.reconcile(mockContext, recorder);
+
+    Assertions.assertEquals(ReconcileProgress.completeAndImmediateRequeue(), progress);
+    Assertions.assertNull(getWorkload());
+    ArgumentCaptor<ApplicationStatus> captor = ArgumentCaptor.forClass(ApplicationStatus.class);
+    verify(recorder).persistStatus(any(), captor.capture());
+    Assertions.assertEquals(
+        ApplicationStateSummary.SchedulingFailure,
+        captor.getValue().getCurrentState().getCurrentStateSummary());
```
peter-toth, line 768: **Finding 2.** This test passes on `d848ee2`, where none of the Kueue code exists. `getDriverPreResourcesSpec()` is left unstubbed, so the reconcile NPEs on the very next line and the same `catch` writes the same `SchedulingFailure` with the same `completeAndImmediateRequeue()`. `assertNull(getWorkload())` holds too, because base never builds one.

Both messages, printed from the captured status:

```
head:         Failed to request driver from scheduler backend. StackTrace: java.lang.UnsupportedOperationException: Kueue does not support SparkApplication with dynamic allocation (spark.dynamicAllocation.enabled=true) yet.
base d848ee2: Failed to request driver from scheduler backend. StackTrace: java.lang.NullPointerException: Cannot invoke "Object.getClass()" because "meta" is null
```

Asserting the message closes it. I ran this on both refs: it passes at head and fails on base.

```suggestion
        ApplicationStateSummary.SchedulingFailure,
        captor.getValue().getCurrentState().getCurrentStateSummary());
    Assertions.assertTrue(
        captor.getValue().getCurrentState().getMessage().contains("dynamic allocation"),
        captor.getValue().getCurrentState().getMessage());
```

The other three that pass on base are fine as they are: `suspendedAppWithQueueNameDoesNotCreateKueueWorkload`, `driverRequestedBeforeBypassesKueueAdmission` and the cluster's `masterRequestedBeforeBypassesKueueAdmission` pin an *ordering*, so they fail on the mutation that matters (moving the check above the gate) even though base has no check to move.

peter-toth (commented):

This is the piece that makes the Kueue work real: both init steps request admission and hold resource creation until Kueue grants quota, with `ADMITTED` / `PENDING` / `STALE` mapped onto proceed, default requeue, and a 5s requeue. The informer filters are the right call — `isAdmissionChanged` keeps Kueue's status churn on a pending `Workload` out of the per-resource reconcile budget. I also confirmed the invariant the `STALE` branch rests on: building the same cluster's `Workload` four times gives one hash, even though the `SPARK_MASTER_OPTS` string inside it comes out in scrambled conf order, so a pending `Workload` cannot be deleted and recreated on every reconcile.

Two things need fixing. Suspending a resource that is *already* queued leaves its `Workload` in place with `active: true`, so Kueue admits it and holds the quota while the operator refuses to start the driver. And `unsupportedKueueSpecFailsScheduling` passes on `d848ee2` for an entirely unrelated reason, so it does not pin what it claims to.

### Blocking
- **1.** *Suspending an already-queued resource holds the quota, and `active: !suspend` is unreachable:* the admission block sits below the suspend gate, so `buildWorkload` is only ever reached with `suspend == false` and nothing ever writes `active: false`. A resource suspended while pending keeps an `active: true` `Workload`. [inline: `AppInitStep.java:107`](https://github.com/apache/spark-kubernetes-operator/pull/843#discussion_r4038881565)
- **2.** *`unsupportedKueueSpecFailsScheduling` green-lights on base:* on `d848ee2` it passes with `java.lang.NullPointerException: ... "meta" is null` from an unstubbed mock, not from the `UnsupportedOperationException`. Asserting the message makes it fail there. [inline: `AppInitStepTest.java:740`](https://github.com/apache/spark-kubernetes-operator/pull/843#discussion_r4038881577)

### Non-blocking
- **3.** *A Kueue API failure permanently fails a `SparkCluster`:* `requestAdmission`'s exceptions land in the `catch (Exception)` that writes `SchedulingFailure`, and for a cluster that becomes `Failed` on the next reconcile with no way back. [inline: `ClusterInitStep.java:85`](https://github.com/apache/spark-kubernetes-operator/pull/843#discussion_r4038881591)
- **4.** *An eviction now wakes the operator and is then discarded:* `isAdmissionChanged` fires a reconcile when an admitted `Workload` stops being admitted, but no step reads it, so the driver and executors keep running after Kueue has taken the quota back. The unsupported list is the place to say so. [inline: `spark_custom_resources.md:605`](https://github.com/apache/spark-kubernetes-operator/pull/843#discussion_r4038881603)
- **5.** *The `SparkCluster` request paragraph reads as a promise, not a floor:* with neither `SPARK_WORKER_CORES` nor a CPU limit set, Spark's worker advertises the node's core count, so the 1-CPU default is far below what the pod will use. [inline: `spark_custom_resources.md:601`](https://github.com/apache/spark-kubernetes-operator/pull/843#discussion_r4038881613)
- **6.** *The reconcile-budget cost of queueing is answered only in the test data:* the comment here describes a production property, but the response is making the e2e's SparkPi run long enough to outlast the window. A short real job pays the same cost. [inline: `spark-example-queued.yaml:31`](https://github.com/apache/spark-kubernetes-operator/pull/843#discussion_r4038881621)

spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/ClusterInitStep.java
```diff
@@ -77,6 +78,19 @@ public ReconcileProgress reconcile(
       }
     }
     try {
+      // Like the suspend hold, a master requested before must complete its initialization.
+      if (KueueWorkloadFactory.hasQueueName(cluster) && !isMasterRequested(context)) {
+        AdmissionResult admission =
+            KueueWorkloadUtils.requestAdmission(
+                context.getClient(), KueueWorkloadFactory.buildWorkload(cluster));
```
peter-toth: **Finding 3.** `requestAdmission` throws for reasons that have nothing to do with the spec: `IllegalStateException` when the `Workload` can neither be read nor created, and a `KubernetesClientException` propagated from the stale-`Workload` delete, which it deliberately does not swallow "so that the resource is not created until the stale Workload is gone" (documented at `KueueWorkloadUtils.java:80-86`, and see [r4029146888](https://github.com/apache/spark-kubernetes-operator/pull/835#discussion_r4029146888)).

Inside this `try`, both become `SchedulingFailure`. For a `SparkCluster` that is the end: `SparkClusterReconciler.getReconcileSteps` maps `SchedulingFailure` to `ClusterUnknownStateStep`, which appends `Failed` unconditionally, and `ClusterInitStep` is never selected again. Observed with a client whose `Workload` calls fail:

```
### progress = ReconcileProgress(completed=true, requeue=true, requeueAfterDuration=PT0S)
### state    = SchedulingFailure
### message  = Failed to request Spark cluster from scheduler backend. StackTrace: java.lang.IllegalStateException: Failed to request Kueue Workload with name: sparkcluster-cluster1
```

So a `SparkCluster` submitted while the operator is missing the Kueue RBAC verb, or while Kueue is briefly unreachable, is permanently dead and has to be recreated. A `SparkApplication` at least gets the restart policy.

Note that `ReconcilerUtils.getResource` swallows every non-404 `KubernetesClientException` and reports the resource as absent, so a 403 on the GET reaches this path as "create it", which is what produced the message above.

The gate is not resource creation, so it wants its own error handling: keep `buildWorkload` inside the `try` so an unsupported spec still lands in `SchedulingFailure`, and treat an API failure as a retry.

```java
      if (KueueWorkloadFactory.hasQueueName(cluster) && !isMasterRequested(context)) {
        Workload desired = KueueWorkloadFactory.buildWorkload(cluster);
        AdmissionResult admission;
        try {
          admission = KueueWorkloadUtils.requestAdmission(context.getClient(), desired);
        } catch (IllegalStateException | KubernetesClientException e) {
          log.warn("Failed to request Kueue admission, will retry.", e);
          return completeAndRequeueAfter(KueueWorkloadUtils.STALE_WORKLOAD_REQUEUE_INTERVAL);
        }
        ...
```

`AppInitStep.java:108` has the same shape.

docs/spark_custom_resources.md
```diff
@@ -557,9 +557,52 @@ spec:
...
+  `Workload` so that Kueue releases the quota. A restarted attempt is queued again. Resources
+  retained by `resourceRetainPolicy` keep the quota until they are released.
+* A `SparkCluster` requests the resources set on the `master` and `worker` containers of its pod
+  templates. A missing request defaults to the limit, or else to 1 CPU and `SPARK_DAEMON_MEMORY`
+  plus overhead. A worker uses `SPARK_WORKER_CORES` for the CPU and adds `SPARK_WORKER_MEMORY` to
+  the memory when they are set. A `SparkCluster` keeps the quota until it is deleted.
+* `spec.suspend` takes precedence. A suspended resource does not get a `Workload` at all.
+* Dynamic allocation, a `SparkCluster` with `minWorkers < maxWorkers`, and pod template files set
+  through `spark.kubernetes.{driver,executor}.podTemplateFile` are not supported yet. Such a
+  resource fails with `SchedulingFailure` instead of being queued.
```
peter-toth, line 614: **Finding 4.** The list is missing the limitation a Kueue user is most likely to assume works. `isAdmissionChanged` is symmetric, so an admitted `Workload` that stops being admitted does fire a reconciliation — the operator is woken and then throws the signal away. `AppInitStep.reconcile` returns `proceed()` at `:69` once the state is past initializing, the Kueue block is additionally skipped by `isDriverRequested`, and no other step or observer reads `Workload` status. So when Kueue evicts — preemption by a higher `WorkloadPriorityClass`, `spec.active` set to `false`, the `ClusterQueue` stopped — it releases the quota and the driver and executors keep running. Real usage then exceeds the quota with nothing to correct it.

This was left open deliberately at the `KueueWorkloadUtils` stage, on the grounds that `AdmissionResult` cannot express "was admitted, then evicted" and that deciding it needs an observer rather than the init steps. This PR is the one that wires admission into reconciliation and the one that adds the wake-up, so it is the right place to state the gap even if the fix is a follow-up.

```suggestion
  resource fails with `SchedulingFailure` instead of being queued.
* Preemption is not honored yet. The operator checks the admission only before it creates the
  resources, so a later eviction (or `spec.active` set to `false` on the `Workload`) releases the
  Kueue quota while the driver and executors, or the master and workers, keep running.
```

docs/spark_custom_resources.md
```diff
@@ -557,9 +557,52 @@ spec:
...
+  API server, so `kubectl get` shows an empty `Current State` and no events are published until the
+  `Workload` is admitted. Use `kubectl get workload` to see the admission status. If the spec
+  changes while waiting, the `Workload` is recreated with the new resource requests.
+* When a `SparkApplication` attempt stops and its resources are released, the operator deletes the
+  `Workload` so that Kueue releases the quota. A restarted attempt is queued again. Resources
+  retained by `resourceRetainPolicy` keep the quota until they are released.
+* A `SparkCluster` requests the resources set on the `master` and `worker` containers of its pod
+  templates. A missing request defaults to the limit, or else to 1 CPU and `SPARK_DAEMON_MEMORY`
+  plus overhead. A worker uses `SPARK_WORKER_CORES` for the CPU and adds `SPARK_WORKER_MEMORY` to
+  the memory when they are set. A `SparkCluster` keeps the quota until it is deleted.
```
peter-toth: **Finding 5.** Everything here is accurate about what the operator computes, which is what makes it misleading as guidance: it does not say that the fallback is a floor rather than the pod's footprint. `WorkerArguments.inferDefaultCores()` is `Runtime.getRuntime.availableProcessors()` and `inferDefaultMemory()` is `max(totalMb - 1024, 1024)`, and with no cpu or memory limit on the container there is no cgroup bound for the JVM to read. So a worker with neither `SPARK_WORKER_CORES` nor a CPU limit advertises the whole node to its executors while the `Workload` asks for 1 CPU.

This is the finding I left on the PR that added the calculation ([r4036382704](https://github.com/apache/spark-kubernetes-operator/pull/841#discussion_r4036382704)), and this section is the first user-facing place it can be said.

```suggestion
  the memory when they are set. A `SparkCluster` keeps the quota until it is deleted. Set a cpu and
  memory request or limit on the `worker` container, or `SPARK_WORKER_CORES` and
  `SPARK_WORKER_MEMORY`: a worker with none of them advertises the whole node to its executors,
  well above the default the `Workload` requests.
```

tests/e2e/kueue/spark-example-queued.yaml
```diff
@@ -0,0 +1,37 @@
...
+  namespace: default
+  labels:
+    kueue.x-k8s.io/queue-name: ($LOCAL_QUEUE)
+spec:
+  mainClass: "org.apache.spark.examples.SparkPi"
+  jars: "local:///opt/spark/examples/jars/spark-examples.jar"
+  # The Kueue admission takes one more reconciliation, which uses up the per-resource rate limit
+  # (5 reconciliations per 15 seconds) before the driver starts. Run longer than that window so
+  # that the driver transitions are observed after the rate limit is refreshed.
+  driverArgs: ["10000"]
```
peter-toth, line 33: **Finding 6.** The comment is right, and the numbers check out: `maxLoopForPeriod` is 5 and `refreshPeriodSeconds` is 15 (`docs/config_properties.md:50-51`). But what it describes is a property of every queued resource, not of this test. The reconciliation that creates the `Workload` and returns `PENDING` is a token a non-queued application never spends, so a short queued job reaches the limiter one event earlier and its terminal transition can be observed up to a refresh period late.

Making SparkPi run 10000 iterations moves the e2e past the window, which is a reasonable way to keep the test deterministic. It also means the suite no longer covers the case the comment is about. Two things worth doing:

- Say it in the user-facing section, next to the sentence about the `Workload` informer, so an operator running short queued jobs is not surprised by a late `Succeeded`:

  ```
  * A queued resource spends one reconciliation on the admission itself, so with the default
    rate limiter (5 loops per 15s) a short application may report its terminal state up to one
    refresh period later than an unqueued one.
  ```

- Reword the comment here to say the test is avoiding the window rather than implying the cost is a test artifact.

The `withOnAddFilter(workload -> false)` / `withOnUpdateFilter(isAdmissionChanged)` pair already removes the larger part of this. I checked that the filters do not affect the informer's cache: `InformerEventSource.onAddOrUpdate` updates `PrimaryToSecondaryIndex` and `TemporaryResourceCache` before it consults either filter, so only event propagation is suppressed.

dongjoon-hyun:
Thank you, @peter-toth . I will revise this PR later because I'm away due to my PTO today.

dongjoon-hyun:
All tests passed.

spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppInitStep.java
dongjoon-hyun: Thank you for the reproduction. I applied the smallest fix in 34e083c. When `spec.suspend` holds a queued resource, `AppInitStep` and `ClusterInitStep` now delete its `Workload` via `KueueWorkloadUtils.releaseWorkload`, so it no longer keeps the quota. The resource is queued again when it is resumed.

- `suspendingQueuedAppReleasesKueueWorkload` and `suspendingQueuedClusterReleasesKueueWorkload` cover it.
- `suspendedAppWithQueueNameDoesNotCreateKueueWorkload` now stubs the client instead of asserting `never().getClient()`.
- The `Kueue` section of the docs describes the new behavior.

I'd like to leave the `spec.active` approach, which keeps the queue position across a suspend, as a follow-up. Until then, `.active(!suspend)` in `KueueWorkloadFactory` remains unreachable.

spark-operator/src/test/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppInitStepTest.java
dongjoon-hyun: Good catch. I added the message assertion as you suggested in 34e083c.

spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/ClusterInitStep.java
dongjoon-hyun: Agreed. In 34e083c, `buildWorkload` stays inside the `try`, so an unsupported spec still ends up in `SchedulingFailure`. An `IllegalStateException` or `KubernetesClientException` from `requestAdmission` is logged and retried after `STALE_WORKLOAD_REQUEUE_INTERVAL` instead. I applied the same to `AppInitStep`, and `kueueApiFailureIsRetried` in both step tests covers it.

For now, the retry is visible only in the operator log. I'd like to surface it as a Kubernetes `Warning` event in a follow-up.

docs/spark_custom_resources.md
dongjoon-hyun: Added your suggested bullet to the `Kueue` section in 34e083c. Honoring an eviction needs an observer beyond the init steps, so I'll handle it in a follow-up.

docs/spark_custom_resources.md
dongjoon-hyun: Added the guidance to the `SparkCluster` bullet in 34e083c, based on your suggestion.

tests/e2e/kueue/spark-example-queued.yaml
dongjoon-hyun: Thank you for checking that the filters do not affect the informer cache. In 34e083c, I added a bullet about the extra reconciliation to the `Kueue` section and reworded the comment here to say that the test avoids the window. The underlying rate limit issue is not specific to Kueue, so I'm looking into it separately.

dongjoon-hyun:
Thank you for the review, @peter-toth. I addressed all comments in 34e083c, and CI passed including the `kueue` E2E test.

peter-toth (approved):

dongjoon-hyun:

dongjoon-hyun:
Merged to main

PR #844, opened 2026-09-18
[SPARK-59578] Clarify that Kueue `integrations.externalFrameworks` is not needed in `operations.md`

peter-toth:

This PR replaces the instruction to register `SparkApplication.v1.spark.apache.org` and `SparkCluster.v1.spark.apache.org` in Kueue's `integrations.externalFrameworks` in the Kueue entry of `docs/operations.md` with a sentence saying it is not needed, and why.

The registration has no effect for this operator, so the instruction asks users to configure something that does nothing.

Traced against Kueue `v0.19.4`:

- `integrations.externalFrameworks` only populates a lookup table. `registerExternal` (`pkg/controller/jobframework/integrationmanager.go:149-169`) stores a `PartialObjectMetadata` for the kind in `m.externalIntegrations`, and `setupControllers` (`pkg/controller/jobframework/setup.go:66-70`) calls `RegisterExternalJobType` for external frameworks and nothing else — no controller and no webhook is built for them.
- The only owner-related consumer of that table is `getJobTypeForOwner` (`integrationmanager.go:230-244`).
- Its caller `IsOwnerManagedByKueueForObject` (`integrationmanager.go:358-363`) starts with `metav1.GetControllerOf(obj)`, which returns the owner reference whose `controller` field is true.
- The operator does not set a controller owner reference on the pods it creates: `ModelUtils.buildOwnerReferenceTo` sets `blockOwnerDeletion` only, and the single place the operator sets `controller: true` is `KueueWorkloadFactory`, on the `Workload` itself.

So `GetControllerOf` returns nil for a driver or executor pod and the registry is never consulted. The operator also does not need Kueue to manage `Workload`s on its behalf: it builds them with `KueueWorkloadFactory` and deletes them in `AppCleanUpStep`.

The same reasoning rules out the one case where the registration might have looked load-bearing. `defaultLocalQueueApplies` (`pkg/controller/jobframework/defaults.go:80-88`) has no feature gate, so in a namespace holding a `LocalQueue` named `default` Kueue can add `queue-name: default` to an object whose owner it does not manage. But that branch returns `!IsOwnerManagedByKueueForObject(jobObj)`, which evaluates the same way with or without the registration, again because there is no controller reference to resolve.

No, documentation only.

Manual review against the Kueue `v0.19.4` sources cited above. `integrations.externalFrameworks` had no other mention in the repository, so nothing else goes stale.

Generated-by: Claude Opus 5


dongjoon-hyun (approved):
+1, LGTM. Thank you, @peter-toth .

peter-toth:
Thank you @dongjoon-hyun.

PR #846, opened 2026-09-18
[SPARK-59637] Remove duplicate driver termination check from `AppDriverStartObserver`

dongjoon-hyun:

This PR makes `AppDriverStartObserver` observe the driver start only. The driver termination is observed by `AppDriverReadyObserver`, which always runs with it in the same step. The `driverReady` parameter of `BaseAppDriverObserver.observeDriverTermination` becomes redundant and is removed.

When the operator first sees a driver pod that has already completed in the `DriverRequested` state, both observers report a terminal state. As a result, a succeeded application records a spurious `Failed` state before `Succeeded`, with the message `Driver container is terminated without SparkContext / SparkSession initialization.` This is a false positive: the application initialized its `SparkContext` and completed successfully. The operator cannot tell from a completed pod whether the `SparkContext` was initialized, because the driver has no readiness probe and a completed pod always reports `Ready=False`.

This happens with a short application that is rate-limited, such as a Kueue-queued one, or when the operator restarts.

Only the state transition history and the events are fixed. The final state is unchanged. A succeeded application no longer publishes the spurious `Failed` Warning event.

- Before: `DriverRequested -> Failed -> Succeeded`
- After: `DriverRequested -> Succeeded`

Pass the CIs with the newly added test cases in `AppResourceObserveStepTest` and `SparkAppReconcilerTest`.

Generated-by: Claude Opus 5


dongjoon-hyun:
Could you review this PR when you have some time, @viirya ~?

spark-operator/src/test/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppResourceObserveStepTest.java
```diff
@@ -0,0 +1,119 @@
...
+            .endTerminated()
+            .endState()
+            .endContainerStatus()
+            .endStatus()
+            .build();
+    when(josdkContext.getSecondaryResourcesAsStream(Pod.class))
+        .thenAnswer(invocation -> List.of(driver).stream());
+    int previousSize = app.getStatus().getStateTransitionHistory().size();
+
+    new AppResourceObserveStep(List.of(new AppDriverStartObserver(), new AppDriverReadyObserver()))
```
viirya, line 111: Optional: could we also add a test exercising the observer composition selected by `SparkAppReconciler`? This test constructs the start/ready pair directly, so it would still pass if production stopped pairing them. Since the start observer now relies on the ready observer to handle termination in the same step, covering that wiring would help protect this fix.

viirya (approved):
The change looks correct to me. I checked that `AppDriverStartObserver` and `AppDriverReadyObserver` run together for both `DriverRequested` and `DriverStarted`, while termination after readiness remains handled by `AppDriverRunningObserver`. Removing the duplicate observation preserves the final state used by cleanup and restart handling.

The new tests appropriately check the transition history, which catches the original issue even though the final state was already correct.

Two small suggestions for the PR description:

- This also removes the spurious `Failed` Warning Event: `StatusRecorder` publishes events for newly appended transitions. The user-visible improvement therefore extends beyond the state history.
- Rather than calling `driverReady` “unused,” it would be more precise to say it becomes redundant after removing the start observer’s termination check. It still controlled a branch before this change.

Neither is blocking. I also left an optional test suggestion inline.

dongjoon-hyun:
Thank you for the review, @viirya. I updated the PR description to mention the Warning event and the redundant `driverReady` parameter. I also added a test case in `SparkAppReconcilerTest` (4262b56) that runs the observe steps selected by `getReconcileSteps` for `DriverRequested` and `DriverStarted`, and checks that a completed driver is recorded only once as `Succeeded`.

dongjoon-hyun:
Merged to main

PR #847, opened 2026-09-18
[SPARK-59638] Support K8s events for `Kueue` admission of `SparkApplication` and `SparkCluster`

dongjoon-hyun:

This PR aims to support K8s events for `Kueue` admission of `SparkApplication` and `SparkCluster`.

| Type | Reason | When |
|---|---|---|
| `Normal` | `KueueAdmissionPending` | The `Workload` waits for the admission |
| `Normal` | `KueueAdmitted` | Kueue admitted the `Workload` |
| `Warning` | `KueueAdmissionRequestFailed` | The admission request fails, except for transport-level errors |

`KueueAdmissionPending` is republished while the `Workload` waits. The event sink keys the `Event` on the reason, so a repeat bumps the `count` of the one `Event` instead of creating another.

A persistent admission request failure, such as a missing Kueue or the RBAC rules for it, is now requeued with the default interval instead of 5 seconds, so that its warning is not rewritten every few seconds until a user fixes the cause. A transport-level failure keeps the 5-second retry and publishes no event, like a failed status update.

In addition,

- `holdForKueueAdmission` moves from both init steps into `KueueWorkloadUtils.holdForAdmission(BaseContext<?>, Workload, String)`, so that `AppInitStep` and `ClusterInitStep` keep only their own guards.
- `EventUtils.normal(...)` is added next to `warn(...)`, so that the call sites do not name `EventType` themselves.
- `ReconcilerUtils.isTransientError` becomes `public`.
- The description of `spark.kubernetes.operator.events.enabled` covers the new events.

A resource waiting for Kueue admission has no state transition, so users cannot see why it is stuck via `kubectl describe`. Like `spec.suspend`, the initial `Submitted` status of the first attempt is not persisted, so `kubectl get` shows an empty `Current State` as well.

The events are therefore the only signal a queued first attempt has. Publishing the pending one once per `Workload` would not be enough: Kubernetes retains events only for a limited time (one hour by default), so a resource queued behind a full `ClusterQueue` for longer would end up with neither a status nor an event, and an operator restart during the wait would publish nothing at all because the `Workload` already exists by then.

Yes. When `spark.kubernetes.operator.events.enabled` is `true`, the new events are published for resources labeled with `kueue.x-k8s.io/queue-name`.

Pass the CIs with the newly added unit tests in `AppInitStepTest`, `ClusterInitStepTest`, `KueueWorkloadUtilsTest` and `EventUtilsTest`, and the updated `kueue` E2E test which asserts the `KueueAdmissionPending` and `KueueAdmitted` events for both `SparkApplication` and `SparkCluster`.

Generated-by: Claude Opus 5


dongjoon-hyun:
All tests passed.

spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppInitStep.java
```diff
@@ -186,6 +191,14 @@ private Optional<ReconcileProgress> holdForKueueAdmission(
       admission = KueueWorkloadUtils.requestAdmission(context.getClient(), desired);
     } catch (IllegalStateException | KubernetesClientException e) {
       log.warn("Failed to request Kueue admission, will retry.", e);
+      // Like a status update failure, a transport level failure is not published, since writing
+      // an event would only add load to an API server that is often the cause of the failure.
+      if (!(e instanceof KubernetesClientException kce && ReconcilerUtils.isTransientError(kce))) {
+        EventUtils.warn(
```
peter-toth, line 197: **Finding 1.** This publishes the warning on every retry of the same failure, and the retry runs every 5 seconds, so a persistent failure is an unbounded event-write loop.

I ran 12 reconciles of one `AppInitStep` against a client whose `create()` always returns 403: the recorder got 12 `record` calls. Each of those is not one API call. `DefaultEventSink.emit` GETs the `Event` by name and then creates or patches it, so it is two. At the 5 s requeue that is about 24 `Event` API calls a minute per stuck resource, and the rate limiter does not clamp it (3 loops per 15 s against the 5-per-15-s default).

The failures that reach here are persistent, not momentary:

- `operatorRbac.kueue.enabled` defaults to `false`, so a resource labeled with a queue name on a chart-default install gets 403 on the `Workload`.
- Kueue not installed at all gives 404 on the create.

Neither is transient by `isTransientError`, so both publish. The resource stays in its initializing state, since the first attempt's `Submitted` is never persisted, so nothing ends the loop but a user fixing the RBAC or deleting the resource. Compare `ReconcileError` two rows above in the same doc table, which publishes only on the first attempt of a failure episode.

The 5 s interval is also off-label here. Its own javadoc scopes it to `STALE`: *"It is short because the stale Workload goes away shortly, while an unchanged admission is watched with the default interval."* A failed request is the opposite case. Requeueing it at the default interval fixes the cadence and the event volume together, 120 s instead of 5 s:

```java
      return Optional.of(completeAndDefaultRequeue());
```

Same for `ClusterInitStep.java:200-201`. `docs/configuration.md:77` then needs its "It is retried every 5 seconds" updated too.

docs/configuration.md
```diff
@@ -74,6 +74,15 @@ In addition, the operator publishes the following `Warning` events.
...
 | `CleanupError` | A cleanup throws, so the resource cannot finish deleting. |
 | `StatusUpdateFailed` | A status patch is rejected. Transport-level errors are skipped. |
+| `KueueAdmissionRequestFailed` | Creating, reading or deleting a stale Kueue `Workload` fails. It is retried every 5 seconds. Transport-level errors are skipped. |
+
+For a resource queued by [Kueue](spark_custom_resources.md#kueue), the operator also publishes the
+following `Normal` events, since the resource stays in its initializing state while it waits.
+
+| Reason | When |
+|---|---|
+| `KueueAdmissionPending` | The Kueue `Workload` is created and waits for the admission. It is not published again for each periodic reconcile while the `Workload` waits. |
```
peter-toth, line 84: **Finding 2.** "It is not published again for each periodic reconcile" is the part I want to push back on, because of what the closing note four lines below says: *"the resource status remains the source of truth"*. For a Kueue-queued first attempt the status is exactly what is missing. `spark_custom_resources.md:592` in this PR says `kubectl get` shows an empty `Current State`, because that attempt's `Submitted` is never persisted.

So the single `KueueAdmissionPending` event is the only signal there is, and the API server drops it after `--event-ttl` (one hour by default). A resource queued behind a full `ClusterQueue` for longer than that ends up with no status and no event, which is the state the "Why are the changes needed?" section describes.

There is a second hole with the same cause. When the operator restarts while a resource is queued, `requestAdmission` reads an existing `Workload` and returns `PENDING`, so nothing is ever published for that resource.

Finding 4 is the simplest way out, since it makes the republish free of new state.

spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/ClusterInitStep.java
```diff
@@ -184,17 +189,42 @@ private Optional<ReconcileProgress> holdForKueueAdmission(
...
           completeAndRequeueAfter(KueueWorkloadUtils.STALE_WORKLOAD_REQUEUE_INTERVAL));
     }
     if (admission == AdmissionResult.STALE) {
       return Optional.of(
           completeAndRequeueAfter(KueueWorkloadUtils.STALE_WORKLOAD_REQUEUE_INTERVAL));
     }
-    if (admission == AdmissionResult.PENDING) {
+    String workloadName = desired.getMetadata().getName();
+    if (admission == AdmissionResult.QUEUED) {
+      EventUtils.record(
```
peter-toth, line 209: **Finding 3.** With this PR `holdForKueueAdmission` is about 50 lines here and the same 50 in `AppInitStep.java:183-233`, identical apart from two nouns in the messages and the two `log.debug` lines. The new parts are duplicated verbatim, including the transient-error test and its two-line comment.

Both `SparkAppContext` and `SparkClusterContext` extend `BaseContext`, which already exposes `getClient()` and `getEventRecorder()`, so the whole body moves:

```java
// KueueWorkloadUtils
public static Optional<ReconcileProgress> holdForAdmission(
    BaseContext<?> context, Workload desired, String requested) { ... }
```

where `requested` is `"driver"` or `"master and workers"`. Each step then keeps only its own guard:

```java
    if (!KueueWorkloadFactory.hasQueueName(cluster) || isMasterRequested(context)) {
      return Optional.empty();
    }
    return KueueWorkloadUtils.holdForAdmission(
        context, KueueWorkloadFactory.buildWorkload(cluster), "master and workers");
```

A follow-up PR is fine. The point is that the divergence risk grows with each event added to one copy only.

spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java
```diff
@@ -62,6 +62,12 @@ private KueueWorkloadUtils() {}
   public enum AdmissionResult {
     /** Kueue admitted the Workload, so the requested resources can be created. */
     ADMITTED,
+    /**
+     * The Workload has just been created and waits for quota, so the resource creation is held
+     * like {@link #PENDING}. It is reported apart so that callers can tell a new request from an
+     * unchanged one.
+     */
+    QUEUED,
```
peter-toth, line 70: **Finding 4.** Worth weighing whether `QUEUED` needs to exist at all. `DefaultEventSink.emit` already does the deduplication this enum value buys. It GETs the `Event` by a name derived from the resource and the record's `key` (which `EventUtils.record` sets to the reason), patches `count`, `message` and `lastTimestamp` when it is there, and creates it when it is not.

So publishing `KueueAdmissionPending` on every `PENDING` reconcile gives one `Event` object per resource with a climbing `count`, not one object per reconcile. It also means an `Event` the API server has already dropped is re-created on the next reconcile, which is what finding 2 needs.

What it removes: this enum value, the `absent` flag, the extra `client.resource(desired).get()` at line 105, and the `QUEUED || PENDING` branching in both init steps. `requestAdmission` goes back to a single `getOrCreateSecondaryResource` call.

The honest cost is two `Event` API calls per queued resource per reconcile instead of zero. At the 120 s default interval that is one write a minute per queued resource, so 1000 queued resources is roughly 17 calls/s. That may well be the wrong trade for you, in which case finding 2 needs a different answer.

One part of the strict read is worth keeping either way. `ReconcilerUtils.getResource` swallows a non-404 failure and reports the resource as missing, which is a bug independent of the events. It would fit as a strict variant used for the initial lookup in `getOrCreateSecondaryResource` itself, so every secondary resource benefits.

spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppInitStep.java
```diff
@@ -195,10 +208,27 @@ private Optional<ReconcileProgress> holdForKueueAdmission(
           ReconcileProgress.completeAndRequeueAfter(
               KueueWorkloadUtils.STALE_WORKLOAD_REQUEUE_INTERVAL));
     }
-    if (admission == AdmissionResult.PENDING) {
+    String workloadName = desired.getMetadata().getName();
+    if (admission == AdmissionResult.QUEUED) {
+      EventUtils.record(
```
peter-toth, line 213: **Finding 5.** `EventUtils` already has `warn(recorder, reason, message)` for a fixed type, and before this PR the only `record(..., type, ...)` caller was `StatusRecorder.java:179`, where the type is computed by `eventTypeOf(...)`. The four new call sites have a fixed `NORMAL` and pull `io.javaoperatorsdk.operator.api.event.EventType` into both init steps only to name it.

```java
  /** Publishes a normal event about a Spark resource. See {@link #warn}. */
  public static void normal(ResourceEventRecorder recorder, String reason, String message) {
    record(recorder, EventType.NORMAL, reason, message);
  }
```

The import then goes away in both `AppInitStep` and `ClusterInitStep`.

peter-toth (commented):

The PR adds three Kueue admission reasons and a new `AdmissionResult.QUEUED`, so the pending event is written once per `Workload` rather than once per reconcile. The strict read behind that also stops a failed `Workload` read from being taken for a missing one, which is worth having on its own. My blocking point runs the other way. The `Warning` is rewritten on every 5-second retry, so a persistent failure becomes an unbounded event-write loop. The pending event has the mirror problem. It is written exactly once, so it is gone after the roughly one-hour event retention while the resource may still be queued.

### Blocking
- **1.** *`KueueAdmissionRequestFailed` is rewritten on every 5 s retry:* A persistent 403 or 404 publishes the warning on every requeue, forever. Each publish is a GET plus a PATCH on the `Event`, so a resource stuck on missing Kueue RBAC costs about 24 API calls a minute. Requeue a failed request with `completeAndDefaultRequeue()` and keep the 5-second interval for `STALE`. [inline: `spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppInitStep.java:197`](https://github.com/apache/spark-kubernetes-operator/pull/847#discussion_r4052542520)

### Non-blocking
- **2.** *The pending event expires while the resource is still queued:* It is written once and Kubernetes drops it after about an hour, while a queued first attempt has no persisted status to fall back on. An operator restart during the wait publishes nothing at all, because the `Workload` already exists by then. [inline: `docs/configuration.md:84`](https://github.com/apache/spark-kubernetes-operator/pull/847#discussion_r4052542525)
- **3.** *`holdForKueueAdmission` is now ~50 duplicated lines in both init steps:* The two copies differ only in two nouns and two log lines, and this PR duplicates its new parts verbatim. Moving the body into `KueueWorkloadUtils` over `BaseContext` would leave each step with just its own guard. [inline: `spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/ClusterInitStep.java:209`](https://github.com/apache/spark-kubernetes-operator/pull/847#discussion_r4052542529)

### Alternatives
- **4.** *Drop `QUEUED` and publish `KueueAdmissionPending` on every pending reconcile:* `DefaultEventSink` already aggregates by an event name derived from the reason, so this would remove the enum value, the `absent` flag and the extra read, and would fix finding 2. The cost is two `Event` API calls per queued resource per reconcile instead of none. [inline: `spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java:70`](https://github.com/apache/spark-kubernetes-operator/pull/847#discussion_r4052542534)

### Minor
- **5.** *`EventUtils.normal(...)` to match `warn(...)`:* Add the sibling helper so the four new call sites do not name `EventType` themselves. [inline: `spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppInitStep.java:213`](https://github.com/apache/spark-kubernetes-operator/pull/847#discussion_r4052542535)

spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppInitStep.java
dongjoon-hyun: Thank you for catching this, fixed in 265d4e0.

I split the `catch` instead of moving the whole failure path to the default interval. A transport-level failure keeps the 5-second retry and still publishes nothing, since it costs no event write and benefits from the fast recovery. A persistent failure publishes once and returns `completeAndDefaultRequeue()`, so the event is rewritten every 120 seconds instead of every 5.

`docs/configuration.md` and the javadoc of `STALE_WORKLOAD_REQUEUE_INTERVAL` are updated accordingly, and `kueueApiFailureIsRetried` in both step tests now asserts the default requeue.

docs/configuration.md
dongjoon-hyun: You are right, and this is now fixed through finding 4. `KueueAdmissionPending` is republished on every pending reconcile, so it survives the event retention and is published again after an operator restart, where the existing `Workload` used to make it silent.

The row you quoted is rewritten, and `docs/spark_custom_resources.md` now says why the event is republished, next to the note that the queued first attempt has no persisted status.

spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/ClusterInitStep.java
dongjoon-hyun: Done in this PR rather than a follow-up, since the duplication grew with the new events.

The body moved to `KueueWorkloadUtils.holdForAdmission(BaseContext<?>, Workload, String)`, where `requested` is `"driver"` or `"master and workers"`, and each init step keeps only its own guard as you sketched. The two `log.debug` lines are merged into one that takes the `Workload` name and `requested` as parameters.

spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java
dongjoon-hyun: Taking this alternative. `QUEUED`, the `absent` flag and the extra `client.resource(desired).get()` are gone, `requestAdmission` is back to a single `getOrCreateSecondaryResource` call, and `KueueAdmissionPending` is published on every `PENDING` reconcile. The two `Event` API calls per queued resource per reconcile are worth it here, since the queued first attempt has no status to fall back on.

The strict read went away with the enum value. The `ReconcilerUtils.getResource` issue you noted is real but independent, and making the initial lookup in `getOrCreateSecondaryResource` strict changes every secondary resource, so I would rather file it as its own ticket than widen this PR.

spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppInitStep.java
dongjoon-hyun: Added in 265d4e0. `EventUtils.normal(recorder, reason, message)` sits next to `warn(...)`, the four call sites use it, and the `io.javaoperatorsdk.operator.api.event.EventType` import is gone from both `AppInitStep` and `ClusterInitStep`.

dongjoon-hyun:
For `ReconcilerUtils.getResource`, I created an independent PR.
- https://github.com/apache/spark-kubernetes-operator/pull/853

dongjoon-hyun:
Rebased to main to resolve the conflicts.

peter-toth (commented):
Re-checked through [`de626e1`](https://github.com/apache/spark-kubernetes-operator/pull/847/commits/de626e13858d81f72f1ad393a3f9973ad1eda595) — findings 1-5 resolved, nothing new. I mutated each of the four new guards in `holdForAdmission` separately — the transient early-return, the default-interval requeue on a persistent failure, and the two `EventUtils.normal` calls — and the tests kill all of them.

Thanks for working through all of these, @dongjoon-hyun — nothing left open from my side.

peter-toth (approved):

dongjoon-hyun:

dongjoon-hyun:
Merged to main

PR #850, opened 2026-09-18
[SPARK-59654] Resolve Kueue `ResourceFlavor` node labels and tolerations of admitted `Workload`

dongjoon-hyun:

This PR aims to resolve the node labels and tolerations of the Kueue `ResourceFlavor`s assigned to each pod set of an admitted `Workload`.

- Add `KueuePodSetFlavor`, which holds the resolved `nodeSelector` and `tolerations` and applies them to a `PodSpec`.
- Add `KueueWorkloadUtils.resolvePodSetFlavors`, which follows the same rules as Kueue v0.19.4 [`pkg/podset/podset.go`](https://github.com/kubernetes-sigs/kueue/blob/v0.19.4/pkg/podset/podset.go) and [`pkg/util/tolerations/tolerations.go`](https://github.com/kubernetes-sigs/kueue/blob/v0.19.4/pkg/util/tolerations/tolerations.go).

This PR is a preparation step for injecting the admitted flavors into Spark pods, as Kueue built-in integrations do.

No. The new code is not called yet.

Pass the CIs with the newly added test cases.

Generated-by: Claude Opus 5


dongjoon-hyun:
All tests passed.

spark-operator/src/test/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtilsTest.java
```diff
@@ -220,6 +225,124 @@ void releaseWorkloadIgnoresFailures() {
...
+  void resolvePodSetFlavorsMergesFlavorsOfEachPodSet() {
+    Toleration spot = toleration("spot");
+    Toleration gpu = toleration("gpu");
+    createFlavor("cpu-flavor", Map.of("pool", "cpu", "zone", "a"), List.of(spot));
+    createFlavor("gpu-flavor", Map.of("pool", "gpu", "accelerator", "a100"), List.of(spot, gpu));
+    // Like Kueue, a flavor assigned to several resources is applied once
+    Map<String, String> driverFlavors = Map.of("cpu", "cpu-flavor", "memory", "cpu-flavor");
+    // Like Kueue, a later flavor overwrites a node label, which is in the resource name order
+    Map<String, String> executorFlavors =
+        Map.of("cpu", "cpu-flavor", "nvidia.com/gpu", "gpu-flavor");
```
peter-toth, line 238: **Finding 1.** This `Map.of` is what feeds the ordering the assertion checks, and `Map.of`'s iteration order is salted per JVM run, so the test only catches a loss of the sort about half the time.

I removed the `new TreeMap<>(...)` from `resolvePodSetFlavors` (leaving `new LinkedHashSet<>(assignment.getFlavors().values())`) and ran `KueueWorkloadUtilsTest` eight times in fresh JVMs: it failed on 5 and **passed on 3**. A probe printing `Map.of("cpu", ..., "nvidia.com/gpu", ...).keySet()` over six runs gave `[cpu, nvidia.com/gpu]` four times and `[nvidia.com/gpu, cpu]` twice, since `java.util.ImmutableCollections.SALT` is derived from `System.nanoTime()` at class init.

To be clear, this is not a flaky test. The real code sorts, so CI stays green either way. It is the regression-catching power that is a coin flip, and the sort is the PR's one deliberate divergence from Kueue, whose `FromAssignment` ranges over the map and so has no defined order at all.

A `LinkedHashMap` in the reverse of resource-name order pins it. Without the sort the iteration is then gpu-before-cpu, `pool` comes out `cpu`, and the assertion fails on every run:

```java
    Map<String, String> executorFlavors = new LinkedHashMap<>();
    executorFlavors.put("nvidia.com/gpu", "gpu-flavor");
    executorFlavors.put("cpu", "cpu-flavor");
```

`driverFlavors` on line 235 can stay a `Map.of`, since both of its values are the same flavor.

spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java
```diff
@@ -127,6 +137,97 @@ public static AdmissionResult requestAdmission(
     return AdmissionResult.PENDING;
   }
 
+  /**
+   * Resolves the node selector and tolerations of the ResourceFlavors which Kueue assigned to each
+   * pod set of the admitted Workload, in the same way as Kueue built-in integrations. The flavors
```
peter-toth, line 142: **Finding 2.** "in the same way as Kueue built-in integrations" does not hold for Topology Aware Scheduling, and TAS is on by default in the version you ported from.

`podset.FromAssignment` does two more things, before the flavor loop this PR mirrors:

```go
	if features.Enabled(features.TopologyAwareScheduling) && assignment.TopologyAssignment != nil {
		if podSet.TopologyRequest == nil || (...) {
			info.Annotations[kueue.PodSetUnconstrainedTopologyAnnotation] = "true"
		}
		info.SchedulingGates = append(info.SchedulingGates, corev1.PodSchedulingGate{
			Name: kueue.TopologySchedulingGate,
		})
	}
```

`pkg/features/kube_features.go` in v0.19.4 lists `TopologyAwareScheduling: {{0.9, false, Alpha}, {0.14, true, Beta}}`, so it is enabled unless an admin turns it off, and it applies to any `ResourceFlavor` with `spec.topologyName`.

Without the scheduling gate the pods go straight to kube-scheduler, which ignores the topology domains Kueue assigned. That is wrong placement with no error anywhere. The operator also cannot notice today: the curated model has no `topologyName` on `ResourceFlavorSpec` and no `topologyAssignment` on `PodSetAssignment`, so both are invisible here.

Nothing is broken yet, since nothing calls this. Two asks:

- Now: scope the claim, e.g. "resolves the node labels and tolerations like Kueue's `podset.FromAssignment`. The TAS scheduling gate and annotation are not handled."
- Before this gets wired up: fail loudly rather than mis-place. Adding `topologyName` to `ResourceFlavorSpec` and throwing `UnsupportedOperationException` when a resolved flavor has it matches how `KueueWorkloadFactory.buildWorkload` already rejects dynamic allocation, HPA and pod template files, and the init steps already turn that into `SchedulingFailure`.

spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueuePodSetFlavor.java
```diff
@@ -0,0 +1,95 @@
...
+
+  private static final String OPERATOR_EQUAL = "Equal";
+
+  /**
+   * Adds the node selector and tolerations to the given pod spec. A node selector conflict must be
+   * checked beforehand, see {@link KueueWorkloadUtils#resolvePodSetFlavors}.
+   *
+   * @param podSpec The pod spec to be modified in place.
+   */
+  public void applyTo(final PodSpec podSpec) {
```
peter-toth, line 48: **Finding 3.** The precondition the javadoc states is checked against a different object than the one `applyTo` mutates, so a caller that gets it wrong silently overwrites the user's node selector instead of failing.

`resolvePodSetFlavors` passes `podSetNodeSelector(desired, ...)` to `checkNoNodeSelectorConflict`, i.e. the node selector of the **Workload's** pod set template. `applyTo` then merges into whatever `PodSpec` it is handed, and `mergedNodeSelector.putAll(nodeSelector)` on line 53 lets the flavor win unconditionally. Kueue never has that split: `podset.Merge` runs `utilmaps.HaveConflict` on the very `spec.NodeSelector` it is about to overwrite, and returns an error.

For a `SparkApplication` the two are produced by independent code. The Workload's driver pod set template gets its node selector from `KueueWorkloadFactory.decorateNodeSelector` (`spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadFactory.java:526`), a reimplementation of Spark's rules, while the real driver pod spec comes from `SparkAppSubmissionWorker` via Spark's own `BasicDriverFeatureStep`. They agree today. The check would not notice if they ever stopped agreeing.

This record already hosts the toleration half of Kueue's merge (`addTolerations`, `isSameToleration`), so the node-selector half fits beside them. Move `checkNoNodeSelectorConflict` here and call it from both places:

```java
  public void applyTo(final PodSpec podSpec) {
    Map<String, String> mergedNodeSelector = new HashMap<>();
    if (podSpec.getNodeSelector() != null) {
      checkNoNodeSelectorConflict(podSpec.getNodeSelector(), nodeSelector);
      mergedNodeSelector.putAll(podSpec.getNodeSelector());
    }
    mergedNodeSelector.putAll(nodeSelector);
```

The pod set name only feeds the message, so `resolvePodSetFlavors` can keep passing it and the `applyTo` path can use a name-free variant. Keeping the early check in `resolvePodSetFlavors` is still worth it, so a conflict is reported before anything is created.

spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java
```diff
@@ -56,6 +64,8 @@ public final class KueueWorkloadUtils {
    */
   public static final Duration STALE_WORKLOAD_REQUEUE_INTERVAL = Duration.ofSeconds(5);
 
+  private static final int HTTP_NOT_FOUND = 404;
```
peter-toth, line 67: **Finding 4.** The module already takes this from the JDK: `ReconcilerUtils.java:27` and `ProbeService.java:22` both do `import static java.net.HttpURLConnection.HTTP_NOT_FOUND`. `Constants.HTTP_TOO_MANY_REQUESTS` exists only because the JDK has no constant for 429, so a new one for 404 is the odd case out.

```java
import static java.net.HttpURLConnection.HTTP_NOT_FOUND;
```

and drop the field.

peter-toth (commented):

The PR ports Kueue's `podset.FromAssignment` plus the nodeSelector/tolerations half of `podset.Merge` into `resolvePodSetFlavors` and a `KueuePodSetFlavor` record, ahead of the PR that will inject them into Spark pods. I checked it against Kueue v0.19.4 line by line and the merge, dedup and conflict semantics match, down to the `key=/value1=/value2=` message shape of `utilmaps.HaveConflict`. Sorting the flavors by resource name is a deliberate improvement over Kueue's randomized map range, and that is where my blocking point is: the test that pins it only catches a regression on about 60% of JVM runs. The other gap is Topology Aware Scheduling, which Kueue enables by default and whose scheduling gate `FromAssignment` also injects.

### Blocking
- **1.** *The deterministic-flavor-order test is a coin flip:* It drives the ordering through `Map.of`, whose iteration order is salted per JVM run. With the `TreeMap` sort removed, `KueueWorkloadUtilsTest` passed on 3 of 8 runs. A `LinkedHashMap` in reverse resource-name order makes it deterministic. [inline: `spark-operator/src/test/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtilsTest.java:238`](https://github.com/apache/spark-kubernetes-operator/pull/850#discussion_r4052619912)

### Non-blocking
- **2.** *"In the same way as Kueue built-in integrations" misses TAS:* `podset.FromAssignment` also adds the `TopologySchedulingGate` and the unconstrained-topology annotation, and TAS is default-on since Kueue 0.14. Scope the claim now, and reject a `topologyName` flavor before this gets wired up, so a TAS user fails loudly instead of getting silently mis-placed pods. [inline: `spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java:142`](https://github.com/apache/spark-kubernetes-operator/pull/850#discussion_r4052619914)
- **3.** *`applyTo` documents a precondition instead of enforcing it:* The conflict check runs against the Workload's pod set template, while `applyTo` overwrites whatever `PodSpec` the caller passes. Kueue's `podset.Merge` checks the spec it is about to modify; moving `checkNoNodeSelectorConflict` into the record and calling it from `applyTo` too closes the gap. [inline: `spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueuePodSetFlavor.java:48`](https://github.com/apache/spark-kubernetes-operator/pull/850#discussion_r4052619917)

### Minor
- **4.** *Private `HTTP_NOT_FOUND` duplicates the JDK constant:* Two other files in the module static-import `java.net.HttpURLConnection.HTTP_NOT_FOUND`. [inline: `spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java:67`](https://github.com/apache/spark-kubernetes-operator/pull/850#discussion_r4052619919)

spark-operator/src/test/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtilsTest.java
dongjoon-hyun: Thank you for catching this, @peter-toth. `executorFlavors` is now a `LinkedHashMap` in the reverse resource-name order, as you suggested. I verified it the same way you did: with the `TreeMap` sort removed from `resolvePodSetFlavors`, the test now fails on every run (3 of 3 in fresh JVMs) instead of the coin flip. `driverFlavors` stays a `Map.of`, since both of its values are the same flavor.

spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java
dongjoon-hyun: Thank you, fixed. It now uses `import static java.net.HttpURLConnection.HTTP_NOT_FOUND` like `ReconcilerUtils` and `ProbeService`, and the private field is gone.

spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java
dongjoon-hyun: Thank you for the detailed pointer to `podset.FromAssignment`. You are right that Topology Aware Scheduling is not covered here. I would like to handle it in the follow-up which wires these flavors into the pods, so that the `topologyName` field on `ResourceFlavorSpec`, the loud rejection, and the javadoc wording land together in one consistent change rather than a partial one here.

spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueuePodSetFlavor.java
dongjoon-hyun: Good point, the conflict check and the mutation do look at different objects today. Since `applyTo` has no production caller yet, I would like to address this in the same follow-up that wires it into the driver/executor pod specs, where the real `PodSpec` shows up and the check can be validated end to end.

peter-toth (commented):
Re-checked through [`ddd29e6`](https://github.com/apache/spark-kubernetes-operator/pull/850/commits/ddd29e6740051bab6d0589a6ee16d0ca53380b40) — findings 1 and 4 resolved, nothing regressed. I re-ran the ordering mutation six times in fresh JVMs and the `LinkedHashMap` kills it on every run, against 5 of 8 before.

I mutation-tested the rest of the new code this round and found nothing further. Every clause of `isSameToleration`, the empty-operator defaulting, both `applyTo` merge branches, the no-admission early return, the `flavorValue != null` conflict guard and the `podSetNodeSelector` null filter are each pinned by an existing test; the three null checks that do survive removal (`getFlavors()`, `getNodeLabels()`, `getTolerations()`) are unreachable through the API, since those fields carry Lombok `@Builder.Default` values and Kueue marks them `omitempty`.

On finding 3, [the sequencing](https://github.com/apache/spark-kubernetes-operator/pull/850#discussion_r4055233557) makes sense and I have nothing to add — the check really is better validated against the real `PodSpec`. Deferring the `topologyName` field and the loud rejection for finding 2 is fine too, with the one exception below.

One thing for that follow-up, which I found while re-checking: the `TopologySchedulingGate` is not only a wait marker. Kueue's `topologyUngater` is what removes it, and in the same patch it copies the assigned domain's node labels into `pod.Spec.NodeSelector` (`pkg/controller/tas/topology_ungater.go:312-318`). So without the gate those labels are never applied at all and the pod can land outside the domain Kueue reserved, rather than merely being scheduled early.

### Non-blocking
- **2.** *The "same way as Kueue built-in integrations" claim (round 1):* Deferring the field and the rejection is fine, but the javadoc sentence lands on main as a claim that is not true, and scoping it needs none of that machinery. I would still take the one-line edit in this PR. [thread](https://github.com/apache/spark-kubernetes-operator/pull/850#discussion_r4052619914)
- **3.** *`applyTo` documents a precondition instead of enforcing it (round 1):* Deferred, agreed, listed only so it is not lost. I confirmed this round that `KueueWorkloadFactory.decorateNodeSelector` is the only thing in the operator that sets a pod node selector, so the two objects do agree today and this is a drift guard rather than a live bug. [thread](https://github.com/apache/spark-kubernetes-operator/pull/850#discussion_r4052619917)

peter-toth (approved):

dongjoon-hyun:
Thank you, @peter-toth . I updated the java doc. For the `topologyUngater` detail, I will carry that into the PR which wires these flavors into the pods, together with the `topologyName` field and the loud rejection.

dongjoon-hyun:
Merged to main

PR #851, opened 2026-09-18
[SPARK-59655] Fill `priorityClassRef` and `priority` of Kueue `Workload`

dongjoon-hyun:

This PR aims to fill `spec.priorityClassRef` and `spec.priority` of Kueue `Workload`s. The priority is decided like Kueue's built-in integrations ([Kueue v0.19.4](https://github.com/kubernetes-sigs/kueue/releases/tag/v0.19.4) (2026-09-10)), in this order:

1. The `WorkloadPriorityClass` named by the `kueue.x-k8s.io/priority-class` label.
2. The `PriorityClass` of the first pod set that has one: `driver` or `master` first.
3. The `globalDefault` `PriorityClass`.

If the named class does not exist, the admission is retried. If the operator cannot read the classes (`403 Forbidden`), it creates the `Workload` without priority. If the label changes while the `Workload` is pending, the `Workload` is updated in place.

Kueue orders the queue and picks preemption targets by `spec.priority`, and the integration has to fill it. Currently, all Spark `Workload`s have the same priority.

No, `Kueue` feature is not released yet.

Pass the CIs with the newly added unit tests and the new Chainsaw step in `tests/e2e/kueue`.

Generated-by: Claude Opus 5


dongjoon-hyun:
All tests passed.

spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java
```diff
@@ -124,9 +140,100 @@ public static AdmissionResult requestAdmission(
       deleteWorkload(client, workload);
       return AdmissionResult.STALE;
     }
+    WorkloadSpec spec = workload.getSpec();
+    if (desired.getSpec().getPriority() != null
```
peter-toth, line 147: **Finding 1.** This clause is what stops a permission blip from wiping a pending Workload's priority, and nothing pins it.

I deleted just `desired.getSpec().getPriority() != null` and ran the whole `spark-operator` suite: **it stays green**. Then, with a Workload already carrying `WorkloadPriorityClass high` (priority 1000) and the priority-class reads failing 403 on the next reconcile, the stored Workload came back `priority=null, priorityClassRef=null`. So the operator would hand it back to Kueue at the default priority 0 because it briefly could not read the class. The CEL rules allow that write, since presence is only frozen once quota is reserved, so nothing downstream catches it either.

This test fails on that mutant and passes as written:

```java
  @Test
  void pendingWorkloadKeepsItsPriorityWhenThePermissionIsLost() {
    createWorkloadPriorityClass("high", 1000);
    KueueWorkloadUtils.requestAdmission(kubernetesClient, workloadWithPriorityClass("high"));
    Assertions.assertEquals(1000, getWorkload().getSpec().getPriority());

    // The operator loses the RBAC rules while the Workload waits for quota
    KubernetesClient forbiddenClient =
        mock(KubernetesClient.class, withSettings().defaultAnswer(delegatesTo(kubernetesClient)));
    KubernetesClientException forbidden = new KubernetesClientException("forbidden", 403, null);
    doThrow(forbidden).when(forbiddenClient).resources(WorkloadPriorityClass.class);
    doThrow(forbidden).when(forbiddenClient).scheduling();

    Assertions.assertEquals(
        AdmissionResult.PENDING,
        KueueWorkloadUtils.requestAdmission(forbiddenClient, workloadWithPriorityClass("high")));
    Assertions.assertEquals(1000, getWorkload().getSpec().getPriority());
    Assertions.assertEquals("high", getWorkload().getSpec().getPriorityClassRef().getName());
  }
```

It needs `org.mockito.AdditionalAnswers.delegatesTo`, `Mockito.doThrow` and `Mockito.withSettings`. `priorityIsNotSetWithoutPermission` covers the same 403 but only through `setPriority`, so it never reaches this branch.

spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java
```diff
@@ -76,16 +90,18 @@ public enum AdmissionResult {
...
-   * @throws IllegalStateException if the Workload can neither be read nor created.
+   * @throws IllegalStateException if the Workload can neither be read nor created, or if its
+   *     priority class does not exist.
    * @throws KubernetesClientException if a stale Workload cannot be deleted. Unlike {@link
    *     #releaseWorkload}, this is not swallowed so that the resource is not created until the
    *     stale Workload is gone.
    */
   public static AdmissionResult requestAdmission(
       final KubernetesClient client, final Workload desired) {
+    setPriority(client, desired);
```
peter-toth: **Finding 2.** Resolving here means every reconcile of a queued resource pays for it, including the ones where nothing changed.

I counted the mock-server requests for a steady-state `PENDING` reconcile with no priority class anywhere: **2 with this line, 1 without it**. The extra one is a cluster-wide `priorityclasses` LIST, because `getPodPriorityClass` falls through to the `globalDefault` lookup, and that is the default shape for a resource with no priority class at all. On the `STALE` path it repeats every 5 s.

Kueue's own integration gates exactly this, in `jobframework.updateWorkloadPriorities`:

```go
	// Resolving only when at least one workload needs a transition keeps
	// steady-state reconciles from re-resolving and overwriting the
	// otherwise-mutable priority value.
	if len(needsClassChange) == 0 {
		return nil
	}
```

`classifyWorkloadsForPriorityUpdate` decides that by comparing class *names*, which costs nothing. The same gate is available here: `KueueWorkloadFactory` copies the owner's labels onto the Workload (`spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadFactory.java:167`), so the stored Workload carries the same `kueue.x-k8s.io/priority-class` label, and its `priorityClassRef.name` is already in hand. So resolve only when the Workload has to be created, or when the existing `priorityClassRef` name differs from the label. That is also exactly the documented behaviour that "a changed value of the same class does not affect the existing `Workload`".

It does need the read to happen before the create, i.e. splitting `getOrCreateSecondaryResource` into a get and a create in this method.

peter-toth (commented):

The PR ports Kueue's priority resolution: the `kueue.x-k8s.io/priority-class` label first, then the first pod set with a `priorityClassName`, then the lowest `globalDefault` `PriorityClass`. I checked each step against Kueue v0.19.4's `pkg/util/priority` and `jobframework`'s `ExtractPriority`/`extractPriorityFromPodSets`, and the order, the group/kind strings and the lowest-globalDefault tiebreak all match. My blocking point is a missing test: the `getPriority() != null` clause is what keeps a permission blip from wiping a pending Workload's priority, and deleting it leaves the whole suite green. Two smaller gaps are that the priority is re-resolved on every reconcile where Kueue explicitly gates that, and that the quota-reserved guard is stricter than the CRD's own CEL rules.

### Blocking
- **1.** *The `getPriority() != null` clause is load-bearing and untested:* Deleting it wipes a pending Workload's `priorityClassRef` and `priority` when the priority-class read 403s, and the suite still passes. A test that fails on that mutant is in the inline comment. [inline: `spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java:144`](https://github.com/apache/spark-kubernetes-operator/pull/851#discussion_r4052922179)

### Non-blocking
- **2.** *The priority is re-resolved on every reconcile:* I measured 2 API calls per steady-state `PENDING` reconcile against 1 without this call, the extra one a cluster-wide `priorityclasses` LIST. Kueue gates the same work on a free class-name comparison, which is available here because the factory copies the label onto the Workload. [inline: `spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java:104`](https://github.com/apache/spark-kubernetes-operator/pull/851#discussion_r4052922186)
- **3.** *The quota-reserved guard is stricter than Kueue and than the CRD:* The CEL rules freeze the ref's presence, group and kind, and the name only for `scheduling.k8s.io/PriorityClass`. A `WorkloadPriorityClass` name and the priority value stay mutable, and Kueue's `hasSameOrEmptyPriorityClass` relies on that. [inline: `spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java:148`](https://github.com/apache/spark-kubernetes-operator/pull/851#discussion_r4052922191)
- **4.** *A missing priority class stalls the resource invisibly:* It retries every 5 s forever with the reason only in the operator log, and the first attempt's status is never persisted, so `kubectl get` and `kubectl describe` both show nothing. The pod `PriorityClass` case is a permanent error that would otherwise surface as `SchedulingFailure`. [inline: `spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java:181`](https://github.com/apache/spark-kubernetes-operator/pull/851#discussion_r4052922195)

### Minor
- **5.** *Private `HTTP_FORBIDDEN` duplicates the JDK constant:* Four other files in the module static-import it from `java.net.HttpURLConnection`. [inline: `spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java:71`](https://github.com/apache/spark-kubernetes-operator/pull/851#discussion_r4052922199)

spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java
```diff
@@ -124,9 +140,100 @@ public static AdmissionResult requestAdmission(
       deleteWorkload(client, workload);
       return AdmissionResult.STALE;
     }
+    WorkloadSpec spec = workload.getSpec();
+    if (desired.getSpec().getPriority() != null
+        && !Objects.equals(spec.getPriorityClassRef(), desired.getSpec().getPriorityClassRef())
+        && (workload.getStatus() == null || !workload.getStatus().isQuotaReserved())) {
+      // Like Kueue, a changed priority class is applied in place so that the Workload keeps its
+      // position in the queue. Kueue does not allow the change once the quota is reserved.
```
peter-toth: **Finding 3.** "Kueue does not allow the change once the quota is reserved" is narrower than that, and the guard above is correspondingly stricter than Kueue.

The Workload CRD freezes four things while `QuotaReserved` is true (`apis/kueue/v1beta2/workload_types.go:1191-1194`): the *presence* of `priorityClassRef`, its `group`, its `kind`, and its `name` **only** for `scheduling.k8s.io/PriorityClass`. The `name` of a `WorkloadPriorityClass` and the `priority` value both stay mutable, and Kueue relies on that: `hasSameOrEmptyPriorityClass` returns true for two `WorkloadPriorityClass` refs with different names, so `updateWorkloadPriorities` goes ahead and writes.

The part of the guard that matters is real. A Workload that reserved quota with no ref can never be given one, which is the first skip in `classifyWorkloadsForPriorityUpdate`. What it also blocks is raising the `WorkloadPriorityClass` on a Workload waiting behind admission checks, which can take minutes with a `ProvisioningRequest`, and that is the case the label exists for.

Mirroring `hasSameOrEmptyPriorityClass` keeps the safe part:

```java
    if (desired.getSpec().getPriority() != null
        && !Objects.equals(spec.getPriorityClassRef(), desired.getSpec().getPriorityClassRef())
        && (workload.getStatus() == null
            || !workload.getStatus().isQuotaReserved()
            || isSameGroupAndKind(
                spec.getPriorityClassRef(), desired.getSpec().getPriorityClassRef()))) {
```

where the helper requires both refs present with equal `group` and `kind`, and equal `name` too when the group is `scheduling.k8s.io`. `quotaReservedWorkloadKeepsPriorityClass` then needs its two classes to differ in group or kind (or to go from a ref to none) so that it keeps testing the case that really is frozen.

spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java
```diff
@@ -124,9 +140,100 @@ public static AdmissionResult requestAdmission(
...
+    try {
+      Map<String, String> labels = workload.getMetadata().getLabels();
+      String workloadPriorityClassName =
+          labels == null ? null : labels.get(Constants.LABEL_WORKLOAD_PRIORITY_CLASS);
+      if (StringUtils.isNotEmpty(workloadPriorityClassName)) {
+        WorkloadPriorityClass workloadPriorityClass =
+            client.resources(WorkloadPriorityClass.class).withName(workloadPriorityClassName).get();
+        if (workloadPriorityClass == null) {
+          throw new IllegalStateException(
+              "Kueue WorkloadPriorityClass " + workloadPriorityClassName + " is not found.");
```
peter-toth: **Finding 4.** A typo in the label leaves the resource stuck with nothing for the user to look at.

`AppInitStep.holdForKueueAdmission` catches `IllegalStateException`, logs a warning and requeues after 5 s. The first attempt's `Submitted` status is never persisted, so `kubectl get` shows an empty `Current State` and `kubectl describe` shows nothing. The result is a resource that retries forever, every 5 s, with the reason only in the operator log, for a plain user error that will not fix itself.

`docs/spark_custom_resources.md` is honest about it ("The resource keeps waiting without a `Workload` while the named priority class does not exist"), but that does not help someone working out why their application never starts.

The pod `PriorityClass` case at line 226 is the worse half, because it is unambiguously permanent: without Kueue the same spec reaches the driver pod and the Priority admission plugin rejects it, so the user gets a `SchedulingFailure` naming the class. Throwing `UnsupportedOperationException` there would restore that. `KueueWorkloadFactory.buildWorkload` already uses it for an unsupported spec and the init steps turn it into `SchedulingFailure`.

For the `WorkloadPriorityClass` case retrying is the right outcome, since an admin may still create the class, so that one needs a signal rather than a different outcome.

spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java
```diff
@@ -56,6 +65,11 @@ public final class KueueWorkloadUtils {
    */
   public static final Duration STALE_WORKLOAD_REQUEUE_INTERVAL = Duration.ofSeconds(5);
 
+  /** Like Kueue, the priority of a Workload without any priority class. */
+  private static final int DEFAULT_PRIORITY = 0;
+
+  private static final int HTTP_FORBIDDEN = 403;
```
peter-toth: **Finding 5.** The JDK has this one, and the module static-imports it elsewhere: `ReconcilerUtils.java:27` for `HTTP_NOT_FOUND`, `SparkExceptionUtils.java:22` and `StatusRecorder.java:22` for `HTTP_CONFLICT`, and `ReadinessProbe.java:22` already pulls in `HTTP_FORBIDDEN` that way. `Constants.HTTP_TOO_MANY_REQUESTS` exists only because 429 has no JDK constant.

```java
import static java.net.HttpURLConnection.HTTP_FORBIDDEN;
```

and drop the field.

dongjoon-hyun:
Thank you for the detailed review, @peter-toth! I addressed 1, 3 and 5.

- **1.** Added `pendingWorkloadKeepsItsPriorityWithoutPermission`. I confirmed it fails when `desired.getSpec().getPriority() != null` is removed and passes as written.
- **3.** Added `isPriorityClassChangeAllowed`, which mirrors the CEL rules: while quota is reserved, the presence, the group and the kind of the ref are frozen, and so is the name of a `scheduling.k8s.io` `PriorityClass`. A `WorkloadPriorityClass` name change now goes through, so a Workload waiting for its admission checks can still be raised. `quotaReservedWorkloadKeepsPriorityClass` is split into a frozen case and a mutable one.
- **5.** Replaced the private constant with `import static java.net.HttpURLConnection.HTTP_FORBIDDEN`.
- **2.** Agreed that the resolution can be gated on a class-name comparison, but it needs `getOrCreateSecondaryResource` split into a get and a create here, which overlaps with #850. Let me do it in a follow-up ticket once #850 is merged.
- **4.** I would like to keep the current behavior. Kueue returns the `Get` error for both the `WorkloadPriorityClass` and the pod `PriorityClass` and retries with backoff, so failing one of them with `SchedulingFailure` would deviate from the integration this PR mirrors. The missing user-facing signal is real, and I think the Kueue admission events in #847 are the right place for it.

spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java
```diff
@@ -124,9 +143,121 @@ public static AdmissionResult requestAdmission(
...
+   * WorkloadPriorityClass stays mutable, which is what Kueue relies on to raise the priority of a
+   * Workload waiting for its admission checks.
+   */
+  private static boolean isPriorityClassChangeAllowed(
+      final Workload workload, final PriorityClassRef desired) {
+    if (workload.getStatus() == null || !workload.getStatus().isQuotaReserved()) {
+      return true;
+    }
+    PriorityClassRef current = workload.getSpec().getPriorityClassRef();
+    return current != null
```
peter-toth: **Finding 6.** Three of this helper's four conditions have no test, and every mutant survives the whole suite.

I removed each in turn and ran `KueueWorkloadUtilsTest`:

| removed | result |
|---|---|
| the `getGroup()` **and** `getKind()` equality checks | suite green |
| `!SCHEDULING_API_GROUP.equals(desired.getGroup())` | suite green |
| `current != null` | suite green |

Group and kind are not separable, since the type's own CEL rule makes `group == 'scheduling.k8s.io'` imply `kind == 'PriorityClass'`, so either one alone still blocks. `quotaReservedWorkloadKeepsFrozenPriorityClass` only covers the presence freeze in the ref-to-none direction.

All three are reachable. These three tests pass as written and kill all three mutants — I ran both arms:

```java
  @Test
  void quotaReservedWorkloadKeepsItsPodPriorityClassName() {
    createPriorityClass("default-a", 50, true);
    KueueWorkloadUtils.requestAdmission(kubernetesClient, workload("owner-uid-1", 1));
    reserveQuota();
    // The cluster-wide default changed, which is a PriorityClass name change
    kubernetesClient.resource(priorityClass("default-a", 50, false)).update();
    createPriorityClass("default-b", 70, true);

    Assertions.assertEquals(
        AdmissionResult.PENDING,
        KueueWorkloadUtils.requestAdmission(kubernetesClient, workload("owner-uid-1", 1)));
    Assertions.assertEquals("default-a", getWorkload().getSpec().getPriorityClassRef().getName());
    Assertions.assertEquals(50, getWorkload().getSpec().getPriority());
  }

  @Test
  void quotaReservedWorkloadDoesNotSwitchPriorityClassGroup() {
    createPriorityClass("default-a", 50, true);
    createWorkloadPriorityClass("high", 1000);
    KueueWorkloadUtils.requestAdmission(kubernetesClient, workload("owner-uid-1", 1));
    reserveQuota();

    // Adding the label would switch the group and the kind of the ref
    Assertions.assertEquals(
        AdmissionResult.PENDING,
        KueueWorkloadUtils.requestAdmission(kubernetesClient, workloadWithPriorityClass("high")));
    Assertions.assertEquals(
        "scheduling.k8s.io", getWorkload().getSpec().getPriorityClassRef().getGroup());
    Assertions.assertEquals(50, getWorkload().getSpec().getPriority());
  }

  @Test
  void quotaReservedWorkloadWithoutPriorityClassDoesNotGainOne() {
    createWorkloadPriorityClass("high", 1000);
    // The Workload is created while the operator cannot read the classes, so it has no ref
    KueueWorkloadUtils.requestAdmission(forbiddenClient(), workloadWithPriorityClass("high"));
    Assertions.assertNull(getWorkload().getSpec().getPriorityClassRef());
    reserveQuota();

    // The permission is back, but Kueue no longer accepts adding a priority class
    Assertions.assertEquals(
        AdmissionResult.PENDING,
        KueueWorkloadUtils.requestAdmission(kubernetesClient, workloadWithPriorityClass("high")));
    Assertions.assertNull(getWorkload().getSpec().getPriorityClassRef());
  }
```

Two helpers to extract: `priorityClass(name, value, globalDefault)` out of `createPriorityClass`, and `forbiddenClient()` out of `pendingWorkloadKeepsItsPriorityWithoutPermission`.

The last one matters most. Without `current != null` it does not merely allow a bad write, it throws `NullPointerException` out of `requestAdmission` on `current.getGroup()`. `holdForKueueAdmission` catches only `IllegalStateException` and `KubernetesClientException`, so the resource ends up in `SchedulingFailure`.

One note on reachability while writing these: a pod template's `priorityClassName` change can never reach this helper, because it changes the pod sets hash and so takes the STALE/recreate path. A `scheduling.k8s.io` ref only changes name through a `globalDefault` swap, which is what the first test uses.

spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java
```diff
@@ -124,9 +143,121 @@ public static AdmissionResult requestAdmission(
...
+      if (StringUtils.isNotEmpty(name)) {
+        PriorityClass priorityClass =
+            client.scheduling().v1().priorityClasses().withName(name).get();
+        if (priorityClass == null) {
+          throw new IllegalStateException("PriorityClass " + name + " is not found.");
+        }
+        return priorityClass;
+      }
+    }
+    return client.scheduling().v1().priorityClasses().list().getItems().stream()
```
peter-toth: **Finding 7.** This lookup runs on every reconcile, so a `Workload` with no label and no pod priority class gets re-priced whenever an admin changes the cluster's default. Kueue never does that.

Observed on the mock server, with nothing about the Spark resource changed between the two reconciles:

```
ref after create                   = PriorityClassRef(group=scheduling.k8s.io, kind=PriorityClass, name=default-a) priority=50
ref after the globalDefault change = PriorityClassRef(group=scheduling.k8s.io, kind=PriorityClass, name=default-b) priority=70
```

`classifyWorkloadsForPriorityUpdate` drops such a Workload from the update set outright:

```go
		if !workload.HasNoPriority(wl) && !workload.IsWorkloadPriorityClass(wl) {
			continue
		}
```

`HasNoPriority` is `PriorityClassRef == nil` and `IsWorkloadPriorityClass` requires the `kueue.x-k8s.io` group, so a `scheduling.k8s.io`-backed Workload never has its priority rewritten after creation. The comment above the function says it outright: "a Pod PriorityClass-backed workload does not follow the label at all". The effect here is that Spark `Workload`s would re-rank inside a `ClusterQueue` while every other workload in it holds still, one write each, which is the opposite of what this PR is trying to match.

This also corrects the fix I sketched for finding 2. Kueue's gate has two parts and I only described the second:

1. skip the Workload entirely when its ref is a Pod `PriorityClass`;
2. otherwise compare the label to the ref's name, and resolve only when they differ.

Part 1 is what fixes this finding, and it is also what makes part 2 well-defined — without it a Pod-`PriorityClass`-backed Workload has no label to compare against, so a name comparison would resolve on every reconcile anyway.

docs/spark_custom_resources.md
```diff
@@ -586,6 +586,13 @@ spec:
   See [Optional Prerequisites](operations.md#optional-prerequisites).
 * The `Workload` is named `<lower-cased kind>-<resource name>` and is owned by the Spark resource,
   so it is garbage collected along with it.
+* Like Kueue built-in integrations, the `Workload` gets the priority of the `WorkloadPriorityClass`
+  named by the `kueue.x-k8s.io/priority-class` label. Without the label, the `priorityClassName` of
+  the driver (or master) pod template is used, then the one of the executor (or worker) pod
+  template, and then the `globalDefault` `PriorityClass`. The resource keeps waiting without a
+  `Workload` while the named priority class does not exist. Changing the label before the
```
peter-toth, line 593: **Finding 8.** "before the `Workload` is admitted" is now broader than what `isPriorityClassChangeAllowed` permits.

Once quota is reserved but before admission, two label changes are silently dropped by the presence and group/kind freeze: adding the label to a `Workload` that has no ref at all, and adding it to one backed by a `globalDefault` `PriorityClass`. Both fall inside the window this sentence promises. The previous "while the `Workload` waits for quota" was accurate, so the new wording needs the exception spelled out:

```
  `Workload` while the named priority class does not exist. Changing the label before the
  `Workload` reserves quota updates its priority in place. After that Kueue freezes the
  presence, group and kind of the priority class, so only the name of a
  `WorkloadPriorityClass` still changes. A changed value of the same class does not affect
  the existing `Workload`.
```

peter-toth (commented):
Re-checked through [`de19a11`](https://github.com/apache/spark-kubernetes-operator/pull/851/commits/de19a1165b5f1d39aed6f0112824d9803e63d9d8) — findings 1, 3 and 5 resolved, nothing regressed. I re-ran the finding-1 mutation and `pendingWorkloadKeepsItsPriorityWithoutPermission` kills it now, and `isPriorityClassChangeAllowed` matches the four CEL rules exactly.

On finding 4, [the reasoning](https://github.com/apache/spark-kubernetes-operator/pull/851#issuecomment-5745997189) holds and it is your call, so I am dropping it. One caveat about leaning on #847 for the signal: as written there the warning is republished on every 5-second retry, which is [the blocking point on that PR](https://github.com/apache/spark-kubernetes-operator/pull/847#discussion_r4052542520).

One correction for finding 2's sequencing: the get/create split you need is in #847, not #850. #850 adds `resolvePodSetFlavors` and leaves `requestAdmission`'s get/create untouched.

### Blocking
- **6.** *The new freeze guard's conditions are untested (new):* Three of `isPriorityClassChangeAllowed`'s four conditions survive removal with the whole suite green, and one of them is an NPE guard on a reachable state. Three tests that pass as written and kill all three mutants are in the inline comment. [inline: `spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java:176`](https://github.com/apache/spark-kubernetes-operator/pull/851#discussion_r4057064979)

### Non-blocking
- **2.** *The priority is re-resolved on every reconcile (round 1):* Still 2 API calls per steady-state `PENDING` reconcile against 1 without the call, re-measured on this head. Deferring is fine; see the correction above on which PR carries the split, and the gate shape in finding 7. [inline: [r4052922186](https://github.com/apache/spark-kubernetes-operator/pull/851#discussion_r4052922186)]
- **7.** *A `globalDefault` change re-prices pending `Workload`s (late catch):* I observed a pending Workload move from `default-a`/50 to `default-b`/70 after only the cluster's `globalDefault` flipped. Kueue's `classifyWorkloadsForPriorityUpdate` drops Pod-`PriorityClass`-backed Workloads from the update set outright, so Spark Workloads would re-rank in a `ClusterQueue` while everything else in it holds still. [inline: `spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java:255`](https://github.com/apache/spark-kubernetes-operator/pull/851#discussion_r4057064983)

### Minor
- **8.** *The reworded doc sentence over-promises (new):* "before the `Workload` is admitted" now covers a window in which the new guard blocks the label change. [inline: `docs/spark_custom_resources.md:593`](https://github.com/apache/spark-kubernetes-operator/pull/851#discussion_r4057064988)

dongjoon-hyun:
Thanks again, @peter-toth! Rebased on `main` (which now carries #850) and addressed 6, 7 and 8.

- **6.** Added tests for each condition of `isPriorityClassChangeAllowed`, and I re-ran the mutations: removing the follow rule fails `pendingWorkloadWithPodPriorityClassDoesNotFollowTheLabel` and `pendingWorkloadKeepsItsPodPriorityClassName`, removing `current != null` fails `quotaReservedWorkloadWithoutPriorityClassDoesNotGainOne`, removing the `desired` group check fails `quotaReservedWorkloadDoesNotSwitchPriorityClassGroup`, and removing `desired != null` fails `quotaReservedWorkloadKeepsFrozenPriorityClass`. `priorityClass(...)` and `forbiddenClient()` are extracted as you suggested.
- **7.** Good catch, fixed. `isPriorityClassChangeAllowed` now starts with part 1 of Kueue's gate: a Workload whose ref is a `scheduling.k8s.io` `PriorityClass` is never rewritten, so a `globalDefault` change no longer re-prices a pending Workload. That also simplifies the quota-reserved check, which now only has to accept a `kueue.x-k8s.io` ref on both sides.
- **8.** Reworded to "before the `Workload` reserves quota", and spelled out both the freeze and the fact that a `PriorityClass`-backed `Workload` does not follow the label.
- **2.** Thank you for the correction on the sequencing, I had the wrong PR. I will do the get/create split as a follow-up after #847 lands, with the two-part gate from finding 7.

One note on the rebase: with `resolvePodSetFlavors` from #850 in the same class, PMD now reports `GodClass` for `KueueWorkloadUtils`, so I added `@SuppressWarnings("PMD.GodClass")` like `KueueWorkloadFactory` has. If you would rather see the priority resolution split into its own class, I am happy to do that instead.

spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java
```diff
@@ -228,6 +260,107 @@ private static void checkNoNodeSelectorConflict(
...
+   * of the priorityClassRef. The name of a WorkloadPriorityClass stays mutable, which is what Kueue
+   * relies on to raise the priority of a Workload waiting for its admission checks.
+   */
+  private static boolean isPriorityClassChangeAllowed(
+      final Workload workload, final PriorityClassRef desired) {
+    PriorityClassRef current = workload.getSpec().getPriorityClassRef();
+    if (current != null && !Constants.KUEUE_API_GROUP.equals(current.getGroup())) {
+      return false;
+    }
+    if (workload.getStatus() == null || !workload.getStatus().isQuotaReserved()) {
```
peter-toth: **Finding 9.** I re-ran the per-clause sweep on the rewritten guard. Five of the six clauses are killed — the Pod-`PriorityClass` skip, `current != null`, `desired != null`, the desired-group check, and the caller's `getPriority() != null`. This one survives.

Removing it is not a no-op. Without the short-circuit the guard falls straight to `current != null && …`, so a `Workload` that currently has **no** priority class can never gain one — including while it is merely pending. That is the case a user hits most: a queued resource created without the label, then labelled while it waits for quota. Every existing test starts from a `Workload` that already carries a `WorkloadPriorityClass` ref (`low` → `high`), and `quotaReservedWorkloadWithoutPriorityClassDoesNotGainOne` covers only the negative, quota-reserved side of the same transition.

This passes as written and fails with the clause removed — I ran both:

```java
  @Test
  void pendingWorkloadWithoutPriorityClassFollowsAnAddedLabel() {
    createWorkloadPriorityClass("high", 1000);
    // No label and no pod priority class, so the Workload is created without a ref
    KueueWorkloadUtils.requestAdmission(kubernetesClient, workload("owner-uid-1", 1));
    Assertions.assertNull(getWorkload().getSpec().getPriorityClassRef());

    // The label is added while the Workload waits for quota
    Assertions.assertEquals(
        AdmissionResult.PENDING,
        KueueWorkloadUtils.requestAdmission(kubernetesClient, workloadWithPriorityClass("high")));
    Assertions.assertEquals("high", getWorkload().getSpec().getPriorityClassRef().getName());
    Assertions.assertEquals(1000, getWorkload().getSpec().getPriority());
  }
```

It also documents the nil-to-ref direction, which is currently only asserted in its blocked form.

peter-toth (commented):
Re-checked through [`eae6437`](https://github.com/apache/spark-kubernetes-operator/pull/851/commits/eae64373aa6a58c284248cd6c387866c3f54db26) — findings 6, 7 and 8 resolved, nothing regressed. The rewritten `isPriorityClassChangeAllowed` is equivalent to Kueue's two-part gate and simpler than what it replaces: with the Pod-`PriorityClass` skip in front, the quota-reserved branch only has to accept a `kueue.x-k8s.io` ref on both sides.

On the `GodClass` question — yes, split it. I tried it: moving `setPriority`, `getPodPriorityClass` and `isPriorityClassChangeAllowed` into a `KueueWorkloadPriority` class makes `pmdMain` pass with no suppression at all on either class. Finding 10 has the numbers.

One correction to finding 2's plan: #847 as it now stands **removed** the get/create split along with `QUEUED`, so waiting for it will not deliver one. The follow-up has to introduce the split itself.

### Blocking
- **9.** *The quota-reserved short-circuit is unpinned (new):* Removing `if (status == null || !isQuotaReserved()) return true;` leaves the suite green, and that clause is what lets a pending `Workload` with no priority class pick up a newly added label — the feature's headline case. I wrote the test and confirmed it passes as written and fails on the mutant. [inline: `spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java:278`](https://github.com/apache/spark-kubernetes-operator/pull/851#discussion_r4057760272)

### Non-blocking
- **2.** *The priority is re-resolved on every reconcile (round 1):* Still 2 API calls per steady-state `PENDING` reconcile against 1 without the call, re-measured on this head. Deferring is fine; see the sequencing correction above, and note the gate now needs both parts of finding 7's rule. [thread](https://github.com/apache/spark-kubernetes-operator/pull/851#discussion_r4052922186)
- **10.** *`KueueWorkloadUtils` splits cleanly, so the suppression is avoidable (new):* Answering your question — the class goes from under the threshold on main to `WMC=61, ATFD=97, TCC=3.297%` with this PR, and the three priority methods are the cut that clears it. [inline: `spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java:67`](https://github.com/apache/spark-kubernetes-operator/pull/851#discussion_r4057760275)

spark-operator/src/main/java/org/apache/spark/k8s/operator/kueue/KueueWorkloadUtils.java
```diff
@@ -40,21 +42,29 @@
...
+import org.apache.spark.k8s.operator.kueue.v1beta2.WorkloadPriorityClass;
+import org.apache.spark.k8s.operator.kueue.v1beta2.WorkloadSpec;
 import org.apache.spark.k8s.operator.kueue.v1beta2.WorkloadStatus;
 import org.apache.spark.k8s.operator.utils.ModelUtils;
 import org.apache.spark.k8s.operator.utils.ReconcilerUtils;
+import org.apache.spark.k8s.operator.utils.StringUtils;
 
 /** Utilities to create, check and release Kueue Workloads. */
 @Slf4j
+@SuppressWarnings("PMD.GodClass")
```
peter-toth: **Finding 10.** Answering the question at the end of [your comment](https://github.com/apache/spark-kubernetes-operator/pull/851#issuecomment-5751418533): yes, I would take the split.

The numbers argue for it rather than against. `pmdMain` is green on `main` for this class, and with this PR it reports:

```
KueueWorkloadUtils.java:67: GodClass: Possible God Class (WMC=61, ATFD=97, TCC=3.297%)
```

That is 14 over the threshold of 47, not one or two, and `TCC=3.297%` says the methods share almost no state — the class is now three unrelated jobs: the `Workload` lifecycle (`requestAdmission`, `releaseWorkload`, `isAdmitted`, `hashPodSets`), flavor resolution from #850, and priority resolution from this PR. #847 adds `holdForAdmission` on top of that.

I tried the cut. Moving `setPriority`, `getPodPriorityClass` and `isPriorityClassChangeAllowed` into a `KueueWorkloadPriority` class in the same package, with `requestAdmission` calling `KueueWorkloadPriority.setPriority(...)` and `KueueWorkloadPriority.isPriorityClassChangeAllowed(...)`, makes `pmdMain` pass with **no suppression on either class** — the only fallout was the imports left behind in `KueueWorkloadUtils`, which have to move too. It also gives the priority tests an obvious home, mirroring how #850 put `KueuePodSetFlavor` beside the resolver.

The `KueueWorkloadFactory` precedent is real, so suppressing is not unreasonable. The difference is that this class trips the rule because of the change in front of us, and the cut that fixes it is small and available.

dongjoon-hyun:
Thanks, @peter-toth! Rebased on `main` (which now carries #847) and addressed 9 and 10.

- **10.** Took the split. `setPriority`, `getPodPriorityClass` and `isPriorityClassChangeAllowed` moved to a new `KueueWorkloadPriority` class in the same package, and `pmdMain` passes with no suppression on either class, so the `@SuppressWarnings("PMD.GodClass")` I had added is gone. I kept the priority tests in `KueueWorkloadUtilsTest`, since most of them exercise the behavior through `requestAdmission` and share its helpers; only the direct `setPriority` calls moved to the new class. Happy to split the test class too if you prefer.
- **9.** Added `pendingWorkloadWithoutPriorityClassFollowsAnAddedLabel` as written, and confirmed it fails when the quota-reserved short-circuit is removed.
- **2.** Noted, thank you for catching that #847 removed the split. The follow-up will introduce the get/create split itself and use both parts of the gate.

One rebase note: #847's new `kueueTransientApiFailurePublishesNoEvent` tests needed the same `priorityclasses().list()` stub as the other deep-stub clients in those files, because Mockito cannot deep-stub the generic list. Only the mock setup changed there.

peter-toth (commented):
Re-checked through [`b8deec5`](https://github.com/apache/spark-kubernetes-operator/pull/851/commits/b8deec5d6fcee160349080fbc60167c852b326a9) — findings 9 and 10 resolved, nothing new. The split landed cleanly: `pmdMain` is green with no suppression on either class, and I re-ran the per-clause sweep on `isPriorityClassChangeAllowed` in its new home — all five clauses are killed now, including the quota-reserved short-circuit that survived last round.

On the test-class question: I would leave it as is. The tests that go through `requestAdmission` are `requestAdmission` tests and belong where they are. Only the four that call `setPriority` directly would move, and a second fixture set is not worth it for those.

Also worth recording that #847 landing closed the gap behind my withdrawn finding 4 in practice, not just in principle. I drove a labelled application whose `WorkloadPriorityClass` does not exist: no `Workload` is created, the requeue is 120 s, and the recorder gets `WARNING/KueueAdmissionRequestFailed`. A typo in the label is visible to a user now.

### Non-blocking
- **2.** *The priority is re-resolved on every reconcile (round 1):* Re-measured on this head — 2 API calls per steady-state queued reconcile against 1 without the `setPriority` call. In production #847's pending event adds a GET plus a create-or-patch on top, which this probe does not capture because the recorder is mocked. Deferred to the follow-up, which now has the right plan. [thread](https://github.com/apache/spark-kubernetes-operator/pull/851#discussion_r4052922186)

peter-toth (approved):

dongjoon-hyun:

dongjoon-hyun:
Merged to main

PR #853, opened 2026-09-19
[SPARK-59667] Do not treat a failed read as a missing secondary resource

dongjoon-hyun:

This PR makes the initial lookup of `ReconcilerUtils.getOrCreateSecondaryResource` stop taking a read the API server refused for a missing resource. A new private `getResourceStrictly` rethrows such a read — a 401, a 403, and anything else answered with a refusal — and keeps reporting the resource as absent for the cases the create path settles on its own: a `HTTP_NOT_FOUND`, the transient codes `isTransientError` covers, a 500 and a 429, which the create loop already retries and whose ambiguity its `AlreadyExists` re-read resolves.

The two reads inside the retry loop stay lenient, since a failed read there only means that the create has to be retried. The public `ReconcilerUtils.getResource` keeps its lenient contract and now logs a warning instead of swallowing the failure silently.

`getResource` returned an empty `Optional` for a 404 and then fell through to `return Optional.ofNullable(resource)` for every other code, so "the resource is not there" and "the resource could not be read" were the same value. `getOrCreateSecondaryResource` then went on to create a resource whose state it had never read.

For the statuses the create loop retries, that conflation works out in the end — the create either succeeds or hits `AlreadyExists` and the re-read settles it — so those keep reporting absent. A refusal is different: the create has no more standing than the read did, so the write goes out for nothing and the failure that surfaces names the create rather than the read that actually failed first. Not knowing whether a resource exists is not the same as knowing that it does not.

`ReconcilerUtils` is annotated with `@SuppressWarnings("PMD.GodClass")` because the class already sits at the rule's WMC threshold, so adding any method to it trips `GodClass`.

No.

Pass the CIs.

Generated-by: Claude Opus 5


dongjoon-hyun:
Rebased to main.

spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/ReconcilerUtils.java
```diff
@@ -181,24 +187,54 @@ public static void addOwnerReferenceSecondaryResource(
...
+   */
+  private static <T extends HasMetadata> Optional<T> getResourceStrictly(
+      final KubernetesClient client, final T desired) {
+    try {
+      return Optional.ofNullable(client.resource(desired).get());
     } catch (KubernetesClientException e) {
       if (e.getCode() == HTTP_NOT_FOUND) {
         return Optional.empty();
       }
+      if (isTransientError(e) || e.getCode() == HTTP_INTERNAL_ERROR) {
```
peter-toth: **Finding 1.** 429 lands in neither branch, so it falls through to `throw e`. That makes an API Priority and Fairness throttle fatal on a path this method otherwise treats as retryable.

I drove `AppInitStep.reconcile` with a client whose driver-pod `get()` throws 429 and whose `create()` succeeds:

```
main    9a61e23 : resulting state = DriverRequested
head    7989aba : resulting state = SchedulingFailure
```

`RestartConfig.restartPolicy` defaults to `Never` and `RestartPolicy.attemptRestartOnState(Never, …)` is unconditionally false, so that `SchedulingFailure` is the end of the application. All three `AppInitStep` call sites run inside the `catch (Exception e)` that writes it.

The inconsistency is inside this same method. `getOrCreateSecondaryResource` has a dedicated 429 branch on the create side, and `shouldBackoffBeforeRetry` returns true for both 429 and a present `Retry-After`:

```java
          } else if (e.getCode() == Constants.HTTP_TOO_MANY_REQUESTS) {
            log.debug("Server returned 429 Too Many Requests, will retry with backoff");
```

So the codebase already says 429 means "come back later". Your own javadoc frames the throwing bucket as "the API server refused the read", and a throttle is a deferral rather than a refusal, which is the same argument.

`Constants` is already imported here, so it is one clause. I applied exactly this and re-ran the 429 arm, which then reaches `DriverRequested` like main:

```suggestion
      if (isTransientError(e)
          || e.getCode() == HTTP_INTERNAL_ERROR
          || e.getCode() == Constants.HTTP_TOO_MANY_REQUESTS) {
```

`createsResourceWhenInitialReadDoesNotReachApiServer` should then take 429 in its `@ValueSource`. Its name is already a slight stretch for 500, where the server did answer, and would be more so for 429 — something like `createsResourceWhenInitialReadGetsNoTrustworthyAnswer` would cover all three.

spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/ReconcilerUtils.java
```diff
@@ -181,24 +187,54 @@ public static void addOwnerReferenceSecondaryResource(
...
+   * the request did not reach a healthy API server and the create path, which re-reads on an
+   * AlreadyExists conflict, still resolves the actual state.
+   *
+   * @param client The KubernetesClient.
+   * @param desired The desired state of the resource.
+   * @param <T> The type of the resource, extending HasMetadata.
+   * @return An Optional containing the retrieved resource, or empty if not found or not reachable.
+   * @throws KubernetesClientException if the API server refused the read.
+   */
+  private static <T extends HasMetadata> Optional<T> getResourceStrictly(
```
peter-toth, line 225: **Finding 2.** `private` leaves the one `getResource` caller outside this class with exactly the bug the description argues against.

`spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/ClusterInitStep.java:208` is that caller, unchanged on main and on #847's head:

```java
  private boolean isMasterRequested(SparkClusterContext context) {
    return ReconcilerUtils.getResource(context.getClient(), context.getMasterStatefulSetSpec())
        .isPresent();
  }
```

Both of its call sites read a refused read as "the master was never requested". The one that matters is the suspend branch at `ClusterInitStep.java:71`:

```java
    if (cluster.getSpec().isSuspend() && !isMasterRequested(context)) {
      ...
      if (KueueWorkloadFactory.hasQueueName(cluster)) {
        KueueWorkloadUtils.releaseWorkload(context.getClient(), cluster);
      }
```

So a 403 on the `StatefulSet` read while a suspend is applied deletes the Kueue `Workload` of a cluster whose master is running, dropping quota that is still consumed. The Kueue-hold site at `ClusterInitStep.java:178` takes the same false branch but is idempotent, so the suspend one is the fix.

`ReconcilerUtils` is in `…operator.utils` and `ClusterInitStep` in `…reconciler.reconcilesteps`, so package-private will not reach it — it needs `public`. `isMasterRequested` already runs inside `ClusterInitStep.reconcile`'s `try`, so a throw lands in the existing `SchedulingFailure` path with no new handling. A follow-up is fine if you would rather keep this PR to `ReconcilerUtils`, but the helper has to be reachable for that follow-up to exist.

spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/ReconcilerUtils.java
```diff
@@ -55,6 +55,7 @@
 
 /** Utility class for reconciler operations. */
 @Slf4j
+@SuppressWarnings("PMD.GodClass")
```
peter-toth, line 58: **Finding 3.** The premise checks out. I removed the annotation and ran `:spark-operator:pmdMain` on this head:

```
ReconcilerUtils.java:58: GodClass: Possible God Class (WMC=50, ATFD=25, TCC=10.909%)
```

The default threshold is WMC 47, so base sat at 47 and this PR takes it to 50. The mechanism is fine — `GodClass` is class-scoped, there is nothing narrower to suppress — but the effect is that the rule is off for this class from here on, and the number is already moving in the wrong direction.

`TCC=10.909%` is the part worth acting on: these methods barely share state because the class is four unrelated groups. `toUpdateControl`/`toDeleteControl` convert control objects, `getOrCreateSecondaryResource`/`getResource`/`getResourceStrictly`/`deleteResourceIfExists` are CRUD helpers, `isTransientError`/`shouldBackoffBeforeRetry`/`isFirstAttempt` classify failures, and `clone` is a JSON deep copy with nothing to do with reconciling — `ModelUtils` in `spark-operator-api` already owns the `objectMapper` it uses.

I tried moving `clone` out to see whether that alone gets under the threshold and my quick edit tripped other PMD rules, so I am not claiming a specific refactor fixes the number. The ask is only that the suppression not be the permanent answer: a follow-up ticket to split the class, referenced from the annotation, keeps the signal alive.

peter-toth (commented):

The premise is right: `getResource` conflated "absent" with "unreadable", so `getOrCreateSecondaryResource` could create a resource whose state it had never read. The three-way classification is a good call — 404 is absent, an unreachable server is absent because the create path's AlreadyExists re-read resolves the ambiguity anyway, and only a refused read throws. One status code sits on the wrong side of that line: a 429 throws, and the same method treats 429 as retryable with backoff twenty lines above.

### Blocking
- **1.** *A 429 on the initial read is classified as a refusal:* I ran the driver-pod path on both refs — main reaches `DriverRequested`, this head reaches `SchedulingFailure`, which with the default `restartPolicy: Never` ends the application. 429 belongs in the unreachable bucket; I verified the one-line change restores parity. [inline: `spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/ReconcilerUtils.java:229`](https://github.com/apache/spark-kubernetes-operator/pull/853#discussion_r4057681791)

### Non-blocking
- **2.** *The one external `getResource` caller cannot reach the fix:* `getResourceStrictly` is `private`, so `ClusterInitStep.isMasterRequested` still reads a refused `StatefulSet` read as "master was never requested", which makes a suspended cluster release its Kueue `Workload` while the master runs. [inline: `spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/ReconcilerUtils.java:221`](https://github.com/apache/spark-kubernetes-operator/pull/853#discussion_r4057681801)
- **3.** *A class-wide lint suppression as the permanent answer:* The premise checks out — without it PMD reports `WMC=50` against a threshold of 47 — but suppressing `GodClass` for the class silences the rule for whatever gets added next, and the number is already drifting. [inline: `spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/ReconcilerUtils.java:58`](https://github.com/apache/spark-kubernetes-operator/pull/853#discussion_r4057681812)

spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/ReconcilerUtils.java
dongjoon-hyun: Thanks, applied in c55f3e2 — exactly the clause you suggested.

You are right that the inconsistency was already inside this method: the create side has its own 429 branch and `shouldBackoffBeforeRetry` returns true for it, so the read side rethrowing a throttle contradicted the rest of the file. A deferral is not a refusal.

I took your point about the test name too and went with `createsResourceWhenInitialReadFailsRetriably`, which drops the "reach the API server" claim that was already inaccurate for 500. Happy to switch to your `...GetsNoTrustworthyAnswer` if you prefer it. 429 is in the `@ValueSource`, and the case fails against the previous revision.

While re-checking this branch I also fixed two things in the same commit that the 429 change touched:

- The warn in that branch used to say "Failed to reach the API server to read the resource", which was already wrong for 500 and would have been wrong for 429 as well. It now says the read failed and the resource is treated as absent, and the lenient wrapper's message names the refusal so the two stay distinguishable.
- `keepsRetryPathReadLenientOnFailure` was vacuous: it injected 503 on the second `get()`, but 503 is swallowed by the strict read itself, so the lenient wrapper it claims to pin was never exercised. Replacing `getResource` with `getResourceStrictly` at both retry-path call sites left the whole module green at 300 tests. It now injects 403, and that mutation fails.

spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/ReconcilerUtils.java
dongjoon-hyun: Agreed on the substance — `isMasterRequested` reading a refused `StatefulSet` read as "the master was never requested" is the same conflation this PR removes, and the suspend branch releasing a live master's Kueue `Workload` is the harmful form of it. I will consider swapping it over in a follow-up.

One correction on the justification, because it changes what that follow-up has to do. `isMasterRequested` has two call sites and the `try` starts at `ClusterInitStep.java:87`, below the first of them:

- The suspend branch at `ClusterInitStep.java:71` — the one you say matters, and I agree — is **outside** that `try`. A throw there leaves `reconcile()` altogether and is retried by JOSDK, not written as `SchedulingFailure`.
- The Kueue-hold site at `ClusterInitStep.java:178` is inside it, so a throw there does reach the `catch (Exception e)` at `ClusterInitStep.java:150`. That is not "no new handling", though: the javadoc right above it at `ClusterInitStep.java:169` says *"SchedulingFailure is terminal for a cluster, so an API failure of the admission request is retried instead"*. Making the read strict without restructuring would promote a transient blip at that site to exactly the terminal state that comment exists to avoid.

So the follow-up needs the lookup hoisted out of the `try` and computed once, rather than a drop-in replacement — worth noting since the second site is idempotent and would otherwise look free to change. Keeping `getResourceStrictly` private for now on the same reasoning: I would rather widen the visibility in the change that actually needs it than expose it speculatively here.

spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/ReconcilerUtils.java
dongjoon-hyun: Thanks for measuring it independently — that matches what I got (base at 47, this PR to 50).

I take the point that a class-scoped suppression turns the rule off for whatever is added next. I am leaning the other way, though: dropping `GodClass` from `config/pmd/ruleset.xml` entirely rather than carrying per-class suppressions. The ruleset already excludes `CyclomaticComplexity`, `NPathComplexity`, `CognitiveComplexity` and `TooManyMethods`, so `GodClass` is the last complexity metric standing, and on a final class of static helpers the `TCC` term it depends on is not measuring anything real — these methods have no instance state to share, so the number you quoted is close to what any stateless utility would score.

That said, the split you describe is worth doing on its own merits, independent of what the linter says. Let me think about which way to go and I will follow up.

peter-toth (commented):
Re-checked through [`c55f3e2`](https://github.com/apache/spark-kubernetes-operator/pull/853/commits/c55f3e2c5aac962f5aa9574ebb6b741006c06c42) — finding 1 resolved, nothing regressed. The lenient bucket now matches the create loop's retryable set exactly (transient, 500, 429; 409 is unreachable on a GET), which is a better-grounded rule than the list it replaces. I mutated all six clauses of the classification and the suite kills every one, including the two you fixed unprompted: the `keepsRetryPathReadLenientOnFailure` 503→403 swap really was vacuous before, and with 503 restored it survives making both retry-path reads strict.

**I was wrong on finding 2 and you are right.** `try` starts at `ClusterInitStep.java:87`, so the suspend site at `:71` is outside it and a throw there leaves `reconcile()` for JOSDK to retry, not `SchedulingFailure`. My "no new handling" line was simply incorrect.

One refinement that makes the follow-up smaller than your reply assumes: with the classification you just landed, a strict `isMasterRequested` can no longer promote a *transient* blip at `:178`, because transient, 500 and 429 all return absent rather than throwing. The only throw left is a refused read. So the thing to handle is narrower and different — a 403 on the `StatefulSet` read at `:178` would become `SchedulingFailure`, while a 403 on the `Workload` read a few lines later is caught and retried by `holdForKueueAdmission`. That asymmetry is the real reason to hoist, and it survives your fix. Keeping `getResourceStrictly` private until that change is the right call either way.

On finding 3, your reading of `TCC` is right: on a `final` class of static helpers there are no instance fields to share, so the cohesion term is structurally near zero and carries no information. Worth weighing against it: `WMC=51` and `ATFD=26` still do, and dropping `GodClass` from the ruleset retires it for instance-ful classes too, where all three terms mean something. Your call, and the split stands on its own merits regardless.

### Blocking
- **4.** *The description no longer matches the classification it ships (new):* The "Why" section still names a 500 and a network failure as the problem, and "What changes" says absent is kept for "a failure that did not reach a healthy API server" — but the final code deliberately keeps 500, 429 and transport failures as absent, and 500 and 429 did reach the server. Only the refused class changed. Two sentences, and ASF folds the description into the commit message, so it is the version that persists.

spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/ReconcilerUtils.java
dongjoon-hyun: Your refinement is right and my reply overstated it. With the classification that landed in c55f3e2, transient, 500 and 429 all return absent, so a strict `isMasterRequested` can no longer promote a blip at `:178` to `SchedulingFailure` — the only throw left there is a refused read. My "transient blip becomes terminal" line was true of 7989aba and is not true of the current head.

The asymmetry you describe is the sharper version of the argument: a 403 on the `StatefulSet` read at `:178` would land in the `catch` at `:150` and become terminal, while a 403 on the `Workload` read a few lines later is caught by `holdForKueueAdmission` and retried. Same failure, same reconcile, opposite outcomes. That is what the follow-up has to reconcile, and hoisting the lookup out of the `try` is still the shape of it — narrower than I described, but for a reason that survives the fix rather than one the fix removed.

Keeping `getResourceStrictly` private until then, as we both land on.

spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/ReconcilerUtils.java
dongjoon-hyun: The instance-ful-class argument is the strongest one against dropping the rule, and I had not weighted it properly — removing `GodClass` from the ruleset does retire it everywhere, including the classes where all three terms mean something, to fix a class where one of them does not. That is a worse trade than I implied.

So I am no longer leaning either way with confidence. I will decide it in the follow-up rather than here, where whichever way it goes is a change to the whole repository's lint policy and does not belong in a bug fix. The split stands on its own merits regardless of what the linter ends up saying, which is the part we agree on.

dongjoon-hyun:
Thanks for the re-check, and for catching finding 4 — you are right that the description had drifted from what the code actually does, and it is the version that persists in the commit message.

Updated. "Why" no longer names a 500 or a network failure as the problem, since neither changed behavior: it now says that "not there" and "could not be read" were the same value, that for the statuses the create loop retries the conflation works out anyway (create succeeds, or hits `AlreadyExists` and the re-read settles it), and that a refusal is the case where it does not — the create has no more standing than the read did, so the write goes out for nothing and the error that surfaces names the create rather than the read that failed first. "What changes" drops "a failure that did not reach a healthy API server" and lists the absent bucket as what it is.

One deliberate wording choice there. I wrote "the transient codes `isTransientError` covers" rather than naming network failures, because I am not yet sure that bucket does what its comment says. `isTransientError` treats `0` as the network sentinel, but in kubernetes-client 7.8.0 `OperationSupport.requestException(HttpRequest, Throwable, String)` builds the exception with `iconst_m1`, i.e. code `-1`. If that is the path a connection reset takes, then `case 0` never matches one, and the pre-existing comment at the transient re-read — "GET to avoid duplicate create attempt for timeouts (0)" — would not be doing anything either. I have only confirmed the constant in the bytecode so far, not the end-to-end path, so I did not want to assert the network-failure behavior in a description that becomes the commit message. It is `main` code and predates this PR, so I will chase it separately rather than widen this one; flagging it here in case you have context on when that constructor is used.

peter-toth (approved):

dongjoon-hyun:

dongjoon-hyun:
Merged to main

PR #854, opened 2026-09-20
[SPARK-59677] Support K8s events for suspended `SparkApplication` and `SparkCluster`

dongjoon-hyun:

This PR aims to support K8s events for `SparkApplication` and `SparkCluster` held by `spec.suspend`.

| Type | Reason | When |
|---|---|---|
| `Normal` | `SuspendHeld` | The resource is held by `spec.suspend`, so the driver (or master and worker) is not requested |

Like the `KueueAdmissionPending` event of SPARK-59638, it is republished while the hold lasts. The event sink keys the `Event` on the reason, so a repeat bumps the `count` of the one `Event` and refreshes it instead of creating another. An application held in `ScheduledToRestart` gets the same event, because the status of its previous attempt says that a restart is due, not that the next attempt is withheld. Resuming publishes no event of its own, since the following state transition already reports it.

Unlike the Kueue hold, which ends when quota arrives, this one ends only when a user clears `spec.suspend`. Republishing on every reconcile would therefore cost a read and a write on the API server for as long as a resource is parked, so the hold is requeued every 30 minutes instead, which still refreshes the event well within the one hour that the API server retains events by default. Nothing waits for that interval, since clearing `spec.suspend` arrives as a watch event.

The reason is `SuspendHeld` rather than `Suspended` because state transition events use the state name as both the reason and the dedup key, so the latter would collide with the dedicated `Suspended` state planned in a follow-up of SPARK-59542.

In addition,

- `SuspendUtils.holdForSuspend(BaseContext<?>, String)` holds the policy for both init steps, mirroring how SPARK-59638 extracted `KueueWorkloadUtils.holdForAdmission`. Each step keeps only its own guard.
- Suspending a queued resource releases its Kueue `Workload`, and the message now says so: the `KueueAdmissionPending` event it was queued with has no owner reference and outlives that `Workload`, so it would otherwise contradict the new event until the API server drops it.
- A new `suspend-events` E2E group asserts the event for both resource kinds, and a new `events` CI mode installs the operator with `spark.kubernetes.operator.events.enabled=true` through `tests/e2e/helm/events-config-values.yaml`. The existing `suspend` group keeps running with the shipped defaults.
- The description of `spark.kubernetes.operator.events.enabled` covers the new event.

A suspended resource has no state transition, so users cannot see why it is stuck via `kubectl describe`. The initial `Submitted` status of a resource created with `suspend: true` is not persisted either, so `kubectl get` shows an empty `Current State` as well.

Publishing the event once would not be enough: Kubernetes retains events only for a limited time (one hour by default), so a resource suspended for longer would end up with neither a status nor an event.

Yes. When `spark.kubernetes.operator.events.enabled` is `true`, the new event is published for resources held by `spec.suspend`.

To opt out, add `SuspendHeld` to `spark.kubernetes.operator.events.excludedReasons`.

Pass the CIs with the newly added unit tests in `AppInitStepTest` and `ClusterInitStepTest`, and the new `suspend-events` E2E group which asserts the event for both `SparkApplication` and `SparkCluster`.

Generated-by: Claude Opus 5


spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppInitStep.java
```diff
@@ -70,12 +70,7 @@ public ReconcileProgress reconcile(
     }
     SparkApplication app = context.getResource();
     if (app.getSpec().isSuspend() && !isDriverRequested(context)) {
-      log.debug("Application is suspended, driver resources would not be requested.");
-      if (KueueWorkloadFactory.hasQueueName(app)) {
-        // A resource suspended while queued must not keep holding the Kueue quota.
-        KueueWorkloadUtils.releaseWorkload(context.getClient(), app);
-      }
-      return completeAndDefaultRequeue();
+      return SuspendUtils.holdForSuspend(context, "driver");
```
peter-toth: **Finding 1.** Both guards in front of this hold report "already requested" as `false` when the read that answers it failed, so a resource whose driver or master is live can be held. This PR stretches that hold from the default interval to 30 minutes.

`SparkAppContext.getCurrentAttemptDriverPod()` catches `KubernetesClientException` and returns empty, logging "considering it absent" (`spark-operator/src/main/java/org/apache/spark/k8s/operator/context/SparkAppContext.java:135`). `ClusterInitStep.isMasterRequested` goes through the lenient `ReconcilerUtils.getResource`, which does the same. That is the point I raised on #853 at [r4057681801](https://github.com/apache/spark-kubernetes-operator/pull/853#discussion_r4057681801), deferred there as a follow-up.

Observed on this head. A suspended app whose current-attempt driver pod exists in the API server, with the live verification read answering 503:

```
PROBE-PROGRESS >>> ReconcileProgress(completed=true, requeue=true, requeueAfterDuration=PT30M)
PROBE-EVENT >>> The SparkApplication is suspended by spec.suspend, driver would not be requested. Set spec.suspend to false to resume it.
PROBE-POD-STILL-THERE >>> true
```

Two consequences, both of them new here. The event asserts the driver was not requested while it is running, and the republishing keeps it in `kubectl describe` for half an hour. And the next look moves from `PT2M` on base to `PT30M`, so an app stuck in `Submitted` with a live driver waits 15x longer unless a pod update happens to arrive first. That window is exactly the one `isDriverRequested` exists for: the app is in `Submitted` with a live driver because an API write just failed, so the read failing on the same reconcile is not an independent coincidence.

The shape I would use is to let the two checks report "could not tell" apart from "absent", and on that answer return `completeAndDefaultRequeue()` and publish no event. The app side is self-contained, `getCurrentAttemptDriverPod` can surface the distinction instead of swallowing it. The cluster side needs `getResourceStrictly` to be reachable, which is the #853 follow-up, so it can wait for that.

peter-toth (commented):

A resource held by `spec.suspend` has no state transition, and on a first attempt no persisted status either, so nothing tells a user why it is parked. This adds a `Normal` `SuspendHeld` event from a shared `SuspendUtils.holdForSuspend` that both init steps call, and paces the republishing with a 30-minute requeue rather than the default interval. I verified the aggregation the design rests on: `DefaultEventRecorder` digests the event name from kind, uid, type, reason and key, and `DefaultEventSink` reads that name and patches `count` plus `lastTimestamp` when it already exists, so the repeats do collapse into one Event and a dropped one is recreated. The description matches the code, the `excludedReasons` opt-out really does reach `SuspendHeld`, and `./gradlew build` is green on this head. Nothing blocking from my side.

### Non-blocking
- **1.** *Hold taken on a failed read now lasts 30 minutes:* Both "already requested" checks read a failed read as "absent", so a resource whose driver or master is live can be held. I observed a suspended app publish `SuspendHeld` and requeue at `PT30M` while its driver pod was present in the API server, where base requeued at `PT2M`. [inline: [`AppInitStep.java:73`](https://github.com/apache/spark-kubernetes-operator/pull/854#discussion_r4060593561)]
- **2.** *Message claims a pending event that may never have existed:* The Kueue suffix is appended on the queue-name label alone, so a resource created with `suspend: true` and a queue name is told an earlier `KueueAdmissionPending` event "no longer applies" when none was ever published. `delete()` already distinguishes the two cases, so the suffix can be conditional. [inline: [`SuspendUtils.java:90`](https://github.com/apache/spark-kubernetes-operator/pull/854#discussion_r4060593571)]
- **3.** *The republish interval is hardcoded:* The one hour it is sized against is a kube-apiserver flag, so a cluster with a shorter `--event-ttl` has no knob. It also leaves the republishing, the central claim of the design, unreachable for the `suspend-events` group. [inline: [`SuspendUtils.java:47`](https://github.com/apache/spark-kubernetes-operator/pull/854#discussion_r4060593574)]

### Minor
- **4.** *Nothing pins the documented 30 minutes:* Both test classes derive the expected progress from the constant, so `Duration.ofSeconds(120)` passes the whole `:spark-operator` suite. [inline: [`AppInitStepTest.java:98`](https://github.com/apache/spark-kubernetes-operator/pull/854#discussion_r4060593579)]
- **5.** *Over-long line in the suspend section:* 146 characters where the paragraph wraps at 100. [inline: [`docs/spark_custom_resources.md:542`](https://github.com/apache/spark-kubernetes-operator/pull/854#discussion_r4060593584)]
- **6.** *`configuration.md` limits the exception to a first attempt:* A `ScheduledToRestart` hold is neither a first attempt nor without a status, and it gets the republished event too. [inline: [`docs/configuration.md:91`](https://github.com/apache/spark-kubernetes-operator/pull/854#discussion_r4060593589)]

spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/SuspendUtils.java
```diff
@@ -0,0 +1,98 @@
...
+            + resource.getKind()
+            + " is suspended by spec.suspend, "
+            + requested
+            + " would not be requested. Set spec.suspend to false to resume it.";
+    if (KueueWorkloadFactory.hasQueueName(resource)) {
+      KueueWorkloadUtils.releaseWorkload(context.getClient(), resource);
+      // The pending event of a resource that was queued before stays until the API server drops
+      // it, and the operator may not delete events, so say that it no longer applies rather than
+      // leaving a contradicting pair behind.
+      message +=
```
peter-toth, line 83: **Finding 2.** The suffix is appended on the queue-name label alone, so a resource created with `suspend: true` and a queue name gets it even though it was never queued and no `KueueAdmissionPending` event was ever published.

`suspendedAppWithQueueNameDoesNotCreateKueueWorkload` (`spark-operator/src/test/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppInitStepTest.java:782`) is exactly that case, and it asserts only the reason, so the message goes unchecked. What it produces:

```
The SparkApplication is suspended by spec.suspend, driver would not be requested. Set spec.suspend to false to resume it. It holds no Kueue Workload while suspended, so an earlier KueueAdmissionPending event no longer applies.
```

An application reaching `ScheduledToRestart` is the same case, since `AppCleanUpStep` already released the Workload on the way in.

`delete()` tells the two apart, so the suffix can be conditional. Measured against the mock server with the same Workload name: `[]` when it is absent, `[StatusDetails(... kind=Workload, name=sparkapplication-sparkapp1 ...)]` when it is present. So `releaseWorkload` can report it:

```java
if (KueueWorkloadFactory.hasQueueName(resource)
    && KueueWorkloadUtils.releaseWorkload(context.getClient(), resource)) {
  message += " It holds no Kueue Workload while suspended, so an earlier " + ...;
}
```

and `suspendedAppWithQueueNameDoesNotCreateKueueWorkload` can then assert the full message, the way `suspendedAppDoesNotRequestDriver` already does.

spark-operator/src/main/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/SuspendUtils.java
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+   * Interval at which a suspended resource is reconciled, so that its event is republished well
+   * within the one hour that the API server retains events by default. It is deliberately much
+   * coarser than the steady-state reconcile interval: a suspended resource has nothing to observe,
+   * while each republish costs a read and a write on the API server for as long as the hold lasts.
+   * Nothing waits for this interval, since clearing spec.suspend arrives as a watch event that
+   * reconciles the resource right away.
+   */
+  static final Duration SUSPEND_HOLD_REQUEUE_INTERVAL = Duration.ofMinutes(30);
```
peter-toth: **Finding 3.** The one hour this is sized against is the kube-apiserver `--event-ttl` flag, not a fixed property of Kubernetes. An operator on a cluster that sets it below 30 minutes cannot shorten the republishing, so the event disappears between repeats and the hold stops being visible. `STALE_WORKLOAD_REQUEUE_INTERVAL` is a constant too, but nothing on a cluster can make 5 seconds wrong, while the retention this one races is configurable.

It also costs the E2E its main assertion. Republishing is what the design rests on, and with a 30-minute floor the `suspend-events` group cannot reach `count > 1`, so nothing outside the unit tests exercises it.

A `ConfigOption<Long>` next to `MISSING_DRIVER_REQUEUE_INTERVAL_SECONDS` in `SparkOperatorConf`, defaulting to 1800 and with `enableDynamicOverride(true)`, addresses both: `tests/e2e/helm/events-config-values.yaml` can set it to a few seconds and the suite can assert the count goes above one.

spark-operator/src/test/java/org/apache/spark/k8s/operator/reconciler/reconcilesteps/AppInitStepTest.java
```diff
@@ -94,6 +94,9 @@ class AppInitStepTest {
 
   private final ResourceEventRecorder eventRecorder = mock(ResourceEventRecorder.class);
 
+  private static final ReconcileProgress SUSPEND_HOLD_PROGRESS =
+      ReconcileProgress.completeAndRequeueAfter(SuspendUtils.SUSPEND_HOLD_REQUEUE_INTERVAL);
```
peter-toth: **Finding 4.** This derives the expected progress from the constant under test, so the value itself is not pinned. I set `SUSPEND_HOLD_REQUEUE_INTERVAL = Duration.ofSeconds(120)`, the default interval this deliberately departs from, and the whole `:spark-operator` suite still passed.

Three places state 30 minutes to users: the `SuspendHeld` row in `docs/configuration.md:85`, `docs/spark_custom_resources.md:541`, and the field's own javadoc. All three drift silently today. One literal is enough, here or in `ClusterInitStepTest.java:87`:

```java
  private static final ReconcileProgress SUSPEND_HOLD_PROGRESS =
      ReconcileProgress.completeAndRequeueAfter(Duration.ofMinutes(30));
```

`java.time.Duration` is not imported in either test class yet.

docs/spark_custom_resources.md
```diff
@@ -536,8 +536,12 @@ StatefulSets. Setting it back to `false` resumes the regular lifecycle.
 
 `Submitted` here is the operator's in-memory view. For a valid resource created with
 `suspend: true`, the initial `Submitted` status is not persisted to the API server, so
-`kubectl get` shows an empty `Current State` until initialization resumes. An application held
-later, in `ScheduledToRestart`, keeps the status its previous attempt already wrote.
+`kubectl get` shows an empty `Current State` and no state transition events are published until
+initialization resumes. Instead, the `SuspendHeld` [event](configuration.md#kubernetes-events) is
+published when enabled. Since the status is not there to fall back on, it is republished every 30
+minutes while the hold lasts, so that it outlives the event retention of the API server. An application held later, in `ScheduledToRestart`, keeps
```
peter-toth: **Finding 5.** 146 characters, where every other line of this paragraph is 100 or under. Re-wrapping 542-544 gives:

```
minutes while the hold lasts, so that it outlives the event retention of the API server. An
application held later, in `ScheduledToRestart`, keeps the status its previous attempt already
wrote and gets the same event, since that status says that a restart is due, not that the next
attempt is withheld.
```

docs/configuration.md
```diff
@@ -76,17 +76,20 @@ In addition, the operator publishes the following `Warning` events.
...
 | Reason | When |
 |---|---|
+| `SuspendHeld` | The resource is held by `spec.suspend`, so the driver (or master and worker) is not requested. It is republished every 30 minutes while the hold lasts, so a repeat bumps the `count` of the one event rather than creating another. Suspending a queued resource releases its Kueue `Workload`, and the message says so, since the `KueueAdmissionPending` event it was queued with outlives that `Workload`. |
 | `KueueAdmissionPending` | The Kueue `Workload` waits for the admission. It is republished while it waits, so a repeat bumps the `count` of the one event rather than creating another. |
 | `KueueAdmitted` | Kueue admitted the `Workload`, so the driver (or master and worker) is requested. |
 
 The `reason` values are stable, while the `message` values may change between releases. Note that
 Kubernetes retains events only for a limited time (one hour by default), so the resource status
-remains the source of truth.
+remains the source of truth. The exception is a first attempt held by `spec.suspend` or by Kueue,
```
peter-toth: **Finding 6.** This narrows the exception to a first attempt, but an application held in `ScheduledToRestart` is neither a first attempt nor without a persisted status, and it gets the republished event too. `docs/spark_custom_resources.md:542` and `SuspendUtils`'s javadoc both say so, and `suspendedAppScheduledToRestartDoesNotRequestDriver` asserts the event.

The condition that actually holds is that the status does not report the hold, which is also why the `ScheduledToRestart` case needs the event. Something like "a resource held by `spec.suspend` or by Kueue, whose status does not report the hold" covers both.

peter-toth (approved):

dongjoon-hyun:
Thank you for the thorough review, @peter-toth! All six are addressed in 2da0874.

**1. Hold taken on a failed read.** `SparkAppContext.getCurrentAttemptDriverPodStrictly()` now surfaces a failed verification, and `getCurrentAttemptDriverPod()` keeps its lenient contract by wrapping it, so existing callers are unchanged. The suspend guard in `AppInitStep` reads the strict variant and, when the verification fails, returns `completeAndDefaultRequeue()` without publishing anything: an unknown answer is no longer read as "no driver requested", and no longer stretches the next look to the hold interval. `suspendedAppWithUnverifiableDriverIsNotHeld` covers it. The cluster half waits for the `getResourceStrictly` follow-up on #853, as you suggested.

**2. Message claims a pending event that may never have existed.** `KueueWorkloadUtils.releaseWorkload` now reports whether a `Workload` was actually deleted, and the suffix is conditional on that. A resource created with `suspend: true` and a queue name therefore gets the plain message, which `suspendedAppWithQueueNameDoesNotCreateKueueWorkload` now asserts in full.

**3. Hardcoded republish interval.** Added `spark.kubernetes.operator.reconciler.suspendHoldRequeueIntervalSeconds` (`Long`, default `1800`, dynamic override) next to `MISSING_DRIVER_REQUEUE_INTERVAL_SECONDS`, and its description points out that it must stay below the API server `--event-ttl`. `tests/e2e/helm/events-config-values.yaml` sets it to 10 seconds, so the `suspend-events` group asserts `count > 1` again for both resource kinds and the republishing is back under E2E coverage.

**4. Nothing pins the documented 30 minutes.** Both test classes now build the expected progress from a `Duration.ofMinutes(30)` literal rather than from the value under test. Verified that it bites: setting the option default to 120 seconds fails the suspend-hold tests.

**5.** Re-wrapped, the paragraph is back to 100 or under.

**6.** Reworded to "a resource held by `spec.suspend` or by Kueue, whose status does not report the hold", which covers the `ScheduledToRestart` case as well.

dongjoon-hyun:
Merged to main

PR #855, opened 2026-09-20
[SPARK-59676] Use the current branch version as the default JIRA fix version in `merge_spark_pr.py`

dongjoon-hyun:

This PR aims to use the version of the current branch as the default JIRA fix version for the `main` branch in `merge_spark_pr.py`.

A new helper, `get_current_version()`, reads the version from `build.gradle` (e.g. `version = "1.1.0-SNAPSHOT"` -> `1.1.0`) and the `main` branch resolves its default fix version to `kubernetes-operator-<version>`. If that version does not exist among the unreleased JIRA versions, it falls back to the previous behavior of choosing the latest one.

Previously, the `main` branch always took the first entry of the unreleased JIRA versions sorted in reverse order. When a future version such as `kubernetes-operator-2.0.0` is created on JIRA in advance, the merge script suggests it as the default fix version although the `main` branch is still `1.1.0-SNAPSHOT`. Deriving the default from `build.gradle` keeps the suggested fix version consistent with the branch which is actually being merged into.

No, this is a developer-only script change.

Manual review with the existing doctests:

```
$ python3 -m doctest dev/merge_spark_pr.py -v | tail -n 1
```

And, the resolved default fix version was verified on the `main` branch.

```
$ python3 -c "exec(open('dev/merge_spark_pr.py').read().replace('if __name__ == \"__main__\":', 'if False:')); print(get_current_version())"
1.1.0
```

Generated-by: Claude Opus 5


dongjoon-hyun:
All tests passed.

dongjoon-hyun:
Let me merge this because this is a committer-only script which is irrelevant to the `Apache Spark K8s Operator` itself.

dongjoon-hyun:
Merged to main

PR #860, opened 2026-09-21
[SPARK-59694] Do not treat an unanswered request as a permanent failure

dongjoon-hyun:

This PR makes `ReconcilerUtils.isTransientError` classify the failures fabric8 reports without a
response status, and makes `shouldBackoffBeforeRetry` delay before retrying them.

`KubernetesClientException.getCode()` returns `-1`, not `0`, when no response status was received,
so the existing `case 0` never matched anything the client produces. This bug was introduced by the following.
- #530

https://github.com/apache/spark-kubernetes-operator/blob/957d4bc2919abe7c863cace3f4dd773f0ade5fe3/spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/ReconcilerUtils.java#L306-L308

A new `NO_RESPONSE_CODE` constant carries that value, and a new private `brokeOnTheWayToTheApiServer` decides whether such a
failure left the request unanswered. It is written as the complement of the failures that repeating
cannot change — an unparsable answer, a handshake rejected over a certificate, and a rejection
raised before sending, which carries no cause at all — so it does not depend on the HTTP client in
use. That matters because the pinned client reports a connection the peer closes mid-response with
a type of its own that is not even an `IOException`.

`shouldBackoffBeforeRetry` gains the same code: an unanswered request carries no `Status` and so
can never supply the `Retry-After` hint that is the only other way to earn a delay.

No failure ever reported code `0`, so every failure that did not reach a healthy API server was
treated as a decision by that API server. Verified against the pinned client by pointing a real
`KubernetesClient` at local sockets:

| Failure | `getCode()` | Innermost cause | Classified |
| --- | --- | --- | --- |
| Connection refused | `-1` | `java.net.ConnectException` | transient |
| Connect timeout | `-1` | `io.netty.channel.ConnectTimeoutException` | transient |
| Read timeout | `-1` | `io.vertx.core.impl.NoStackTraceTimeoutException` | transient |
| Connection reset | `-1` | `java.net.SocketException` | transient |
| Peer closed mid-response | `-1` | `io.vertx.core.http.HttpClosedException` | transient |
| DNS failure | `-1` | `java.net.UnknownHostException` | transient |
| Unparsable response body | `-1` | `com.fasterxml.jackson.databind.exc.MismatchedInputException` | permanent |
| Rejected before sending | `-1` | none | permanent |

The misclassification reached every caller:

| Caller | Before | After |
| --- | --- | --- |
| `getOrCreateSecondaryResource` | rethrew on the first attempt, skipping the `GET` that avoids a duplicate create, which `AppInitStep` turns into a terminal `SchedulingFailure` under the default `RestartPolicy.Never` | retries with backoff |
| `getResourceStrictly` | reported an unanswered read as a refusal | reports it as one the create path settles |
| `StatusRecorder.persistStatus` | wrote a `StatusUpdateFailed` event per blip, adding load to the server that caused it | stays quiet |
| `KueueWorkloadUtils.retryAfterRequestFailure` | warning event plus the default requeue interval | the short interval, no event |

The classification also holds in the other direction: an unparsable response and a rejection raised
before sending no longer pass as transient, so they keep their warning event and longer requeue
instead of being retried forever in silence.

Yes. A `SparkApplication` that loses its connection to the API server while requesting driver
resources is now retried with backoff instead of failing on the first attempt, and such failures no
longer emit `Warning` events. Retries stay bounded by
`spark.kubernetes.operator.api.secondaryResourceCreateMaxAttempts` (3 by default), so exhausting
them still ends in `SchedulingFailure`. `SparkCluster` master and worker resources are created
through `serverSideApply()` rather than `getOrCreateSecondaryResource()`, so they are unaffected; a
`SparkCluster` benefits through its status updates and its Kueue admission requests.

Pass the CIs.

Generated-by: Claude Opus 5


dongjoon-hyun:
cc @jiangzho

dongjoon-hyun:
All tests passed.

dongjoon-hyun:
Could you review this when you have some time, @viirya ?

spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/ReconcilerUtils.java
```diff
@@ -297,20 +315,53 @@ public static boolean isFirstAttempt(Context<?> context) {
...
+    // Bounded walk: a cyclic cause chain must not hang the reconciler. The marker is nested, since
+    // fabric8 wraps the failure and its own HTTP client wraps it again.
+    Throwable cause = e.getCause();
+    for (int depth = 0; cause != null && depth < MAX_CAUSE_DEPTH; depth++) {
+      // A handshake that failed over a certificate, a host name or a trust store is the one
+      // answer-less failure that repeating cannot fix, so it is grouped with the unusable answers.
+      if (cause instanceof JsonProcessingException
+          || cause instanceof SSLHandshakeException
+          || cause instanceof SSLPeerUnverifiedException) {
+        return false;
```
viirya, line 366: [P2] Preserve retryability for TLS handshake timeouts

Could we distinguish certificate validation failures from transient handshake failures here? Netty's `SslHandshakeTimeoutException` extends `SSLHandshakeException`, so this condition also rejects a TLS handshake timeout caused by a temporarily unresponsive API server or proxy. `getResourceStrictly()` then propagates the initial GET failure, and the CREATE path skips its retry/backoff branch; `AppInitStep` can still turn that temporary failure into `SchedulingFailure`. Please narrow the permanent-failure check and add a handshake-timeout regression test while retaining coverage for certificate rejection.

Reference: [Netty's exception definition at the pinned version](https://github.com/netty/netty/blob/netty-4.2.18.Final/handler/src/main/java/io/netty/handler/ssl/SslHandshakeTimeoutException.java).

spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/ReconcilerUtils.java
```diff
@@ -297,20 +315,53 @@ public static boolean isFirstAttempt(Context<?> context) {
...
+      }
+      cause = cause.getCause();
+    }
+    return e.getCause() != null;
```
viirya, line 370: [P2] Exclude thread interruption from retryable failures

This fallback also classifies thread interruption as a transient transport failure. In fabric8, `OperationSupport.waitForResult()` restores the interrupt flag and wraps `InterruptedException` in `InterruptedIOException`, which can reach this method inside a code `-1` `KubernetesClientException`. The create path will then issue additional GET/CREATE attempts, while `backoffSleep()` immediately exits because the thread remains interrupted. Could we detect actual interruption in the cause chain and stop retrying it? Please avoid excluding every `InterruptedIOException`, since `SocketTimeoutException` extends that class and should remain retryable.

Reference: [fabric8's interruption handling](https://github.com/fabric8io/kubernetes-client/blob/v7.8.0/kubernetes-client/src/main/java/io/fabric8/kubernetes/client/dsl/internal/OperationSupport.java#L489-L510).

viirya (changes requested):
Thanks for fixing the no-response sentinel and adding backoff for transport failures. Using `-1` matches fabric8, and keeping the existing GET-after-failed-CREATE path makes sense.

I found two classification cases that need attention: TLS handshake timeouts are classified as permanent failures, while thread interruption is classified as a retryable transport failure. Details are in the inline comments.

Could we also align the documentation with the actual scope?

- The Javadoc and configuration description for `API_SECONDARY_RESOURCE_CREATE_MAX_ATTEMPTS` still say retryable errors without `Retry-After`, including network timeouts, are retried immediately. Please update these and the corresponding generated documentation.
- The PR description says this protects SparkCluster master-resource creation, but `ClusterInitStep` creates those resources through direct `serverSideApply()` calls rather than `getOrCreateSecondaryResource()`.
- For SparkApplication, retries remain bounded, and exhaustion can still result in `SchedulingFailure`. The user-facing description should make that limit explicit.

The existing tests cover several useful cases. Please also cover the two cases below, plus code `-1` transport failures through the StatusRecorder and Kueue callers, since this classification controls their events and requeue behavior too.

spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/ReconcilerUtils.java
dongjoon-hyun: Good catch. `SslHandshakeTimeoutException` does extend `SSLHandshakeException`, so the check was rejecting it.

Rather than special-casing the Netty type, I keyed the permanent check on the trust failures themselves, which is the set fabric8 uses in `StandardHttpClient.isTerminalTlsTrustFailure`: `CertificateException`, `CertPathValidatorException`, `CertPathBuilderException` and `SSLPeerUnverifiedException`. A rejected certificate nests one of those (`sun.security.validator.ValidatorException` extends `CertificateException`, `SunCertPathBuilderException` extends `CertPathBuilderException`), while a handshake that merely timed out does not, so it stays retryable. This also stops the check from depending on which client reported the handshake.

Verified against the pinned client: a handshake timeout is now transient, a certificate rejection and a peer-unverified failure stay permanent, and a bare `SSLException` from a mid-connection reset is transient.

Added `classifiesHandshakeTimeoutAsTransient`, and `doesNotClassifyCertificateFailureAsTransient` now nests a `CertificateException` under the handshake so the rejection case stays covered.

spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/ReconcilerUtils.java
dongjoon-hyun: Agreed, and thank you for the warning about `InterruptedIOException` — `SocketTimeoutException` extends it, so excluding the whole class would have made plain read timeouts permanent.

The check now matches `InterruptedException` in the cause chain, which is what `waitForResult()` puts under the `InterruptedIOException` and which a read timeout never carries.

Verified by interrupting a call in flight against the pinned client: code `-1`, cause chain `InterruptedIOException <- InterruptedException`, interrupt flag still set, now classified permanent — while `SocketTimeoutException` stays transient.

Added `doesNotClassifyInterruptionAsTransient`.

dongjoon-hyun:
Thanks for the review, @viirya. All five points were correct; `f6e5936` addresses them. The two classification fixes are answered inline.

**Documentation scope.** Updated the Javadoc and the `description` of `API_SECONDARY_RESOURCE_CREATE_MAX_ATTEMPTS`, and regenerated `docs/config_properties.md`. The text no longer claims that retryable errors without a `Retry-After` hint are retried immediately: an unanswered request carries no response status, so it can never supply that hint and is now backed off, which the description says explicitly.

**SparkCluster.** You are right — `ClusterInitStep` creates master and worker resources through `serverSideApply()` and never calls `getOrCreateSecondaryResource()`, so the create path is unaffected for them. A `SparkCluster` benefits only through its status updates and its Kueue admission requests. The user-facing paragraph now says so, and also states that retries stay bounded by this configuration (3 by default), so exhausting them still ends in `SchedulingFailure`.

**Caller coverage.** Added tests that drive a code `-1` failure through the callers whose events and requeue interval this classification controls:

| Test | Covers |
| --- | --- |
| `StatusRecorderTest.publishesNoEventWhenTheRequestGoesUnanswered` | an unanswered status patch publishes no `StatusUpdateFailed` event |
| `StatusRecorderTest.publishesAnEventWhenTheRequestIsRejectedBeforeItIsSent` | a client-side rejection still publishes one |
| `AppInitStepTest.kueueUnansweredRequestPublishesNoEvent` | an unanswered admission request requeues at the short interval with no event |
| `AppInitStepTest.kueueClientSideRejectionPublishesAnEvent` | a client-side rejection requeues at the default interval with an event |

The status-recorder test points the client at a closed port so the failure is a real unanswered request rather than a hand-built exception.

PTAL when you have a moment.

spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/ReconcilerUtils.java
viirya: Thanks, this addresses the issue. Checking the underlying trust failures preserves retryability for `SslHandshakeTimeoutException` without coupling the production code to Netty. The updated certificate-rejection test also retains coverage for the permanent failure case. Resolved from my side.

spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/ReconcilerUtils.java
viirya: Thanks, this addresses the issue. Matching `InterruptedException` in the cause chain handles fabric8’s interruption wrapping while keeping `SocketTimeoutException` retryable. The added regression test covers the distinction we discussed. Resolved from my side.

viirya (approved):
Thanks for addressing the review comments. I checked `f6e5936` and both classification issues are resolved. The documentation now correctly scopes the SparkCluster behavior and bounded retries, and the added StatusRecorder/Kueue tests cover the event and requeue behavior requested.

I also checked the updated classifier against ten focused cases, including handshake timeout, socket timeout, nested certificate failures, and interruption; the results match the intended behavior.

One small documentation follow-up: the description of `API_SECONDARY_RESOURCE_CREATE_INITIAL_BACKOFF_MILLIS` still lists only `Retry-After`, 409, and 429. Please consider mentioning unanswered transport requests there too, including the generated documentation.

No remaining blocking concerns from me, subject to the updated CI passing.

dongjoon-hyun:
Thanks for the approval, @viirya. The documentation follow-up is in `d240956`.

`API_SECONDARY_RESOURCE_CREATE_INITIAL_BACKOFF_MILLIS` now lists the unanswered request alongside `Retry-After`, 409 and 429, in both the Javadoc and the `description`, and `docs/config_properties.md` is regenerated:

> Backoff is applied whenever the API server response carries a Retry-After hint, on 409 (Conflict) and 429 (Too Many Requests) responses, and on a request the API server never answered, which carries no response status and so can never supply a Retry-After hint.

I checked the neighbouring options for the same stale wording. The only other `Retry-After` mention is in `API_RETRY_ATTEMPT_AFTER_SECONDS`, which describes the status-patch retry interval rather than this backoff policy, so I left it alone.

dongjoon-hyun:
Merged to main

PR #861, opened 2026-09-21
[SPARK-59700] Add a minimum interval between unchanged K8s event repeats

dongjoon-hyun:

Add `spark.kubernetes.operator.events.minIntervalSeconds` (default `300`, dynamically
overridable) and enforce it in `ConfigurableEventRecorder`, the single gate every Kubernetes
Event passes through.

An event is dropped only when the same `(metadata.uid, reason, message)` was already published
within the interval. The `message` is part of the identity because `DefaultEventSink` rewrites
the `message` and the `lastTimestamp` of the existing `Event` on every repeat, so only a repeat
that says exactly what the last one said carries nothing new. The first event of an identity is
always published, so a rare event such as `Failed` is never delayed. The window is fixed rather
than sliding, so an event repeated more often than the interval is still republished. The
tracking map is bounded by an inline sweep that drops entries older than the interval, and is
discarded when the feature is off. No new dependency and no background thread.

An event is only published when a reconciliation emits one, so the effective period is the
interval rounded up to the next repeat. The documented rule is therefore to keep this option
below the interval at which the repeats are emitted
(`spark.kubernetes.operator.reconciler.suspendHoldRequeueIntervalSeconds` for `SuspendHeld`,
`spark.kubernetes.operator.reconciler.intervalSeconds` for `KueueAdmissionPending`), which is
stricter than keeping it below the `--event-ttl` of the API server. At the default, this means
`KueueAdmissionPending` bumps its `count` every 360 seconds instead of every 120, and no other
event is affected: `SuspendHeld` already repeats every 1800 seconds, and the rest change their
`message` between repeats.

`tests/e2e/helm/events-config-values.yaml` sets the option to `0`, since the `suspend-events`
suite asserts that the `count` of an unchanged `SuspendHeld` event rises within 120 seconds.

A global QPS cap is deliberately not added: it would need a drop policy that sacrifices first
occurrences, which is exactly what this change guarantees never happens. Event publish/drop
counters are left to a separate ticket.

`DefaultEventSink` writes each event synchronously on the reconcile thread with a `GET` plus a
`CREATE` or a `PATCH` — two API calls per event — and JOSDK has no equivalent of the client-go
`EventRecorder` spam filter. A repeat whose `message` is unchanged only increments the `count`,
so it pays that cost while telling the user nothing new.

This is a prerequisite for defaulting `spark.kubernetes.operator.events.enabled` to `true`. That
default change is out of scope here.

No. K8s Event feature is not released yet.

Pass the CIs with new unit tests.

Generated-by: Claude Opus 5


dongjoon-hyun:
All tests passed.

docs/configuration.md
```diff
@@ -105,6 +107,45 @@ Since `,` is the separator, an expression cannot contain it, e.g. `{1,3}`. An in
...
+starts over.
+
+An event is only published when a reconciliation emits it, so the effective period is this
+interval **rounded up to the next repeat**, not the interval itself. With the defaults,
+`KueueAdmissionPending` is emitted every 120 seconds
+(`spark.kubernetes.operator.reconciler.intervalSeconds`), so a 300 second interval publishes it
+every 360 seconds, and `SuspendHeld` is emitted every 1800 seconds
+(`spark.kubernetes.operator.reconciler.suspendHoldRequeueIntervalSeconds`), so it is unaffected.
+
+Keep the interval **below the interval at which the repeats themselves are emitted**, i.e. below
```
peter-toth: **Finding 1.** "below both of the two options named above" means below `spark.kubernetes.operator.reconciler.intervalSeconds`, which is 120. The default of this option is 300, so the rule declares the shipped default a misconfiguration, and five lines earlier the same section says that default is fine. A user who follows the rule sets the interval under 120, at which point every repeat is published and the pacing stops applying to `KueueAdmissionPending` — the only event the change affects at defaults, per the PR description.

The example given is sound, but it only supports the rule for `SuspendHeld`. I measured all four cases on the timed recorder, recording a repeat every `R` seconds for one hour and listing the publishes:

```
I=300  R=120   [0, 360, 720, 1080, 1440, 1800, 2160, 2520, 2880, 3240, 3600]
I=2400 R=1800  [0, 3600]
I=1800 R=1800  [0, 1800, 3600]
I=1801 R=1800  [0, 3600]
```

So the hazard is not `I` versus `R`, it is the effective period `ceil(I/R)*R` reaching the TTL. For `SuspendHeld` at `R=1800` that does bind at `I <= 1800`, exactly 15x looser than "below 120", and the last two rows show the boundary is precisely there. For `KueueAdmissionPending` at `R=120` it binds at `I <= 3480`.

`ceil(I/R)*R < I + R`, so one rule covers both without the reader doing arithmetic:

```
Keep `minIntervalSeconds` plus the interval at which the repeats are emitted within the
`--event-ttl` of the API server (one hour by default), i.e. below
`3600 - spark.kubernetes.operator.reconciler.suspendHoldRequeueIntervalSeconds` for `SuspendHeld`
and below `3600 - spark.kubernetes.operator.reconciler.intervalSeconds` for
`KueueAdmissionPending`. With the defaults that is 1800 seconds, so the default 300 leaves plenty
of room. At `minIntervalSeconds=2400` the `SuspendHeld` repeat at 1800 seconds is dropped and the
next one lands at 3600 seconds, exactly the TTL, so the event expires in between and the hold
stops being visible.
```

The same rule is stated in two more places and needs the same correction: `SparkOperatorConf.java:373-374` in the option description, and `docs/config_properties.md` which is generated from it. The reciprocal clause added to `SUSPEND_HOLD_REQUEUE_INTERVAL_SECONDS` ("keep `minIntervalSeconds` below this interval") is fine as it stands, since `I <= R` is what binds for that reason.

tests/e2e/helm/events-config-values.yaml
```diff
@@ -19,3 +19,6 @@ operatorConfiguration:
     # The suspend-events suite asserts that a held resource republishes its SuspendHeld event.
     # The default is 1800 seconds, which no assert can wait for.
     spark.kubernetes.operator.reconciler.suspendHoldRequeueIntervalSeconds=10
+    # The repeat above carries the same message, so the default 300 second minimum interval would
+    # drop every one of them and the count would never rise within the assert timeout.
+    spark.kubernetes.operator.events.minIntervalSeconds=0
```
peter-toth: **Finding 2.** `0` takes the `minInterval <= 0` early return, so `suspend-events` — the only suite that installs the operator with events enabled — exercises none of the new code. The map, the `compute`, the sweep and the interval comparison are all skipped, and the unit tests are the sole coverage.

A small positive value gets the coverage for free, because the pacing is a no-op whenever the interval is below the emit interval. With `suspendHoldRequeueIntervalSeconds=10` already set two lines up, I ran `minIntervalSeconds=5` for 120 seconds of repeats:

```
PROBE I=5 R=10 publishes at >>> [0, 10, 20, 30, 40, 50, 60, 70, 80, 90, 100, 110, 120]
PROBE tracked entries >>> 1
```

Every repeat still publishes, so the `count` rises exactly as it does at `0` and the existing assert is unaffected, while the positive-interval path now runs on a real cluster with a real `uid`. That last part is the bit unit tests cannot cover: `uidOf` reads `context.getPrimaryResource().getMetadata().getUid()`, and every unit test hands it a hand-built `SparkApplication` rather than one that came back from the API server.

```suggestion
    # A small positive value keeps every repeat published, since the pacing is a no-op below the
    # requeue interval above, while still exercising the minimum-interval path on a real resource.
    spark.kubernetes.operator.events.minIntervalSeconds=5
```

spark-operator/src/main/java/org/apache/spark/k8s/operator/utils/ConfigurableEventRecorder.java
```diff
@@ -115,6 +166,107 @@ private static boolean isExcluded(String reason) {
...
+    HasMetadata resource = context.getPrimaryResource();
+    ObjectMeta metadata = resource == null ? null : resource.getMetadata();
+    return metadata == null ? null : metadata.getUid();
+  }
+
+  private static long minIntervalNanos() {
+    Long seconds = KUBERNETES_EVENTS_MIN_INTERVAL_SECONDS.getValue();
+    // An unparseable override falls back to the default, but an override of the literal 'null'
+    // resolves to null, which is treated as no limit rather than throwing out of a reconciliation.
+    return seconds == null ? 0L : TimeUnit.SECONDS.toNanos(seconds);
```
peter-toth, line 267: **Finding 3.** The comment is exactly right, which is why the branch deserves the same test `KUBERNETES_EVENTS_ENABLED` got. I confirmed both halves of it:

```
PROBE-NULL-RESOLVES-TO >>> null
PROBE-GARBAGE-RESOLVES-TO >>> 300
PROBE-NULL-PUBLISHES-BOTH >>> confirmed
```

`ConfigOption.resolveValue` sends a non-primitive type through `objectMapper.readValue`, so `"null"` deserialises to `null` while `"not-a-number"` throws `JsonProcessingException` and falls back to the default.

What makes it worth pinning is that the two options fail in *opposite* directions on the same input: `eventsEnabled()` uses `Boolean.TRUE.equals`, so a `"null"` override stops every event, while here it removes the limit entirely. Both are the right choice for their option, and only one of them is currently a test:

```java
  @Test
  void treatsAMalformedMinIntervalOverrideAsNoLimit() {
    // The option is a boxed Long, so a malformed override can resolve to null. Unlike the enabled
    // flag, which then drops every event, a limit that cannot be read must not drop any.
    SparkOperatorConfManager.INSTANCE.refresh(
        Map.of(
            KUBERNETES_EVENTS_ENABLED.getKey(), "true",
            KUBERNETES_EVENTS_MIN_INTERVAL_SECONDS.getKey(), "null"));
    Context<?> context = contextOf("uid-1");
    EventRecord first = EventRecord.normal("KueueAdmissionPending", "queued");
    EventRecord second = EventRecord.normal("KueueAdmissionPending", "queued");

    timedRecorder.record(first, context);
    timedRecorder.record(second, context);

    verify(delegate).record(first, context);
    verify(delegate).record(second, context);
  }
```

peter-toth (commented):

`DefaultEventSink` writes every event with a GET plus a CREATE or PATCH on the reconcile thread, and a repeat whose `message` is unchanged only bumps the `count`, so it pays two API calls to tell the user nothing. This adds `spark.kubernetes.operator.events.minIntervalSeconds` and enforces it in `ConfigurableEventRecorder`, keyed on `(metadata.uid, reason, message)`, with the window anchored at the last *published* time rather than the last attempt. That shape holds up: the first event of an identity always publishes, a changed message always publishes, and because the window is fixed rather than sliding an event repeated faster than the interval still gets through. I measured the effective periods on the timed recorder and they match the description exactly, `./gradlew build` is green, and the `compute`-per-key plus `compareAndSet` sweep is the right concurrency shape for a recorder shared by every reconcile thread.

One Blocking finding, and it is the documentation rather than the code: the tuning rule this PR states contradicts the default it ships, and following it would switch the pacing off for the only event the change affects at defaults.

### Blocking
- **1.** *The documented tuning rule excludes the shipped default:* "keep it below the interval at which the repeats themselves are emitted ... below both of the two options" means below 120, while the default is 300. Measured: the default is safe (`KueueAdmissionPending` publishes every 360s against a 3600s TTL), and the real bound is `minIntervalSeconds + <emit interval> <= --event-ttl`. [inline: [`docs/configuration.md:139`](https://github.com/apache/spark-kubernetes-operator/pull/861#discussion_r4069330734)]

### Non-blocking
- **2.** *The new gate has no E2E coverage:* `minIntervalSeconds=0` turns the feature off for the only suite that has events enabled. `5` keeps the `count` rising every 10 seconds and does exercise the gate, which I verified. [inline: [`events-config-values.yaml:24`](https://github.com/apache/spark-kubernetes-operator/pull/861#discussion_r4069330746)]

### Minor
- **3.** *The null-override branch is untested:* `KUBERNETES_EVENTS_ENABLED` has `treatsAMalformedOverrideAsDisabled` for exactly this, and the two options fail in opposite directions. [inline: [`ConfigurableEventRecorder.java:267`](https://github.com/apache/spark-kubernetes-operator/pull/861#discussion_r4069330773)]

dongjoon-hyun:
Thanks for the careful review, @peter-toth. I reproduced all four probes and the boundary is exactly where you put it. Addressed in 1669b5b.

**1. Fixed.** You are right on both counts: the rule contradicted the shipped default, and `ceil(I/R)*R < I + R` is the bound that matters. I brute-forced the largest safe `I` for both emit intervals and it lands exactly on `TTL - R`, so the bound is tight rather than merely sufficient:

```
R=120   largest safe I = 3480   TTL-R = 3480
R=1800  largest safe I = 1800   TTL-R = 1800
```

The docs now state that limit with the concrete numbers, and note that the default of 300 leaves plenty of room. Corrected in `docs/configuration.md`, in the option description, and in the regenerated `docs/config_properties.md`.

I also corrected the reciprocal clause on `SUSPEND_HOLD_REQUEUE_INTERVAL_SECONDS`, which you were willing to leave as is. `I <= R` only coincides with the real bound because `TTL - R == R` at the default of 1800; at `suspendHoldRequeueIntervalSeconds=600` the real limit is 3000, not 600. All four places now state the same rule.

**2. Adopted the idea, with `15` rather than `5`.** At `I=5, R=10` the effective period is `ceil(5/10)*10 = 10`, so every repeat publishes and the drop path never executes — the positive-interval path runs, but the interval comparison never returns true:

```
I=5   publishes=[0, 10, 20, 30, ...]  count>1 at t=10s  drops in 120s = 0
I=15  publishes=[0, 20, 40, 60, ...]  count>1 at t=20s  drops in 120s = 6
```

At `15` six repeats are actually dropped inside the 120s assert window while the `count` still rises every 20 seconds, six times inside the timeout. Your point about the real `uid` holds either way, and a slower cluster only makes the assert safer, since any reconcile gap above 15 seconds publishes every repeat.

**3. Added, using your test as written.** I confirmed it pins the branch rather than just covering it: mutating `seconds == null ? 0L` to fall back to the default instead makes `treatsAMalformedMinIntervalOverrideAsNoLimit` fail.

`gradle build -x test` and `gradle :spark-operator:test` are green.

docs/configuration.md
```diff
@@ -82,8 +84,8 @@ events, since the resource stays in its initializing state without a state trans
...
-| `KueueAdmissionPending` | The Kueue `Workload` waits for the admission. It is republished while it waits, so a repeat bumps the `count` of the one event rather than creating another. |
+| `SuspendHeld` | The resource is held by `spec.suspend`, so the driver (or master and worker) is not requested. It is republished every 30 minutes while the hold lasts (`spark.kubernetes.operator.reconciler.suspendHoldRequeueIntervalSeconds`), so a repeat bumps the `count` of the one event rather than creating another. A repeat is also subject to [`minIntervalSeconds`](#event-frequency), which must be kept below the requeue interval. Suspending a queued resource releases its Kueue `Workload`, and the message says so, since the `KueueAdmissionPending` event it was queued with outlives that `Workload`; that repeat carries a new `message`, so it is published even within the interval. |
```
peter-toth: **Finding 4.** "which must be kept below the requeue interval" is the rule this commit replaced everywhere else. `grep` finds it in exactly one place, so it is the fifth site rather than one of the four you counted: the `Event frequency` section, both option descriptions and the two generated `config_properties.md` rows all state the corrected bound.

The argument against it is the one you made for the reciprocal clause in [issuecomment-5778311639](https://github.com/apache/spark-kubernetes-operator/pull/861#issuecomment-5778311639): `I <= R` only coincides with `I <= TTL - R` because `TTL - R == R` at the default of 1800. Lower the requeue interval and the two diverge fast, and it diverges in the direction that makes the row wrong rather than merely conservative.

The repo's own configuration is the counterexample. `tests/e2e/helm/events-config-values.yaml` sets `suspendHoldRequeueIntervalSeconds=10` and `minIntervalSeconds=15`, which this row declares illegal, and it is correct — measured on the timed recorder:

```
I=15 R=10 period=20 first publishes=[0, 20, 40, 60]
```

20 seconds against a 3600 second TTL, with every other repeat dropped, which is exactly the behaviour the E2E now relies on.

Since the row already links to the section that states the rule properly, the shortest fix is to stop restating it:

```suggestion
| `SuspendHeld` | The resource is held by `spec.suspend`, so the driver (or master and worker) is not requested. It is republished every 30 minutes while the hold lasts (`spark.kubernetes.operator.reconciler.suspendHoldRequeueIntervalSeconds`), so a repeat bumps the `count` of the one event rather than creating another. A repeat is also subject to [`minIntervalSeconds`](#event-frequency). Suspending a queued resource releases its Kueue `Workload`, and the message says so, since the `KueueAdmissionPending` event it was queued with outlives that `Workload`; that repeat carries a new `message`, so it is published even within the interval. |
```

peter-toth (commented):
Re-checked through [`1669b5b`](https://github.com/apache/spark-kubernetes-operator/commit/1669b5ba875f0e303817e08a276d0fe7c90ae3d5) — findings 1, 2 and 3 resolved, nothing regressed. I re-measured rather than read: the corrected bounds are exact (`I=1800` publishes every 1800 and `I=1801` every 3600; `I=3480` every 3480 and `I=3481` every 3600), the E2E's `I=15` with `R=10` publishes at 0, 20, 40, 60 so `count > 1` lands at 20 seconds against a 120 second timeout, and mutating `seconds == null ? 0L` to fall back to the default makes `treatsAMalformedMinIntervalOverrideAsNoLimit` fail and nothing else. `15` over my `5` is the better call for the reason you gave: at `5` the interval comparison never returns true, so the drop path would not have run. You were also right to correct the reciprocal clause I waved through — `I <= R` only coincides with the real bound at the default, and I should have caught that when I derived `TTL - R`.

One leftover from finding 1's fix.

### Non-blocking
- **4.** *The `SuspendHeld` table row still carries the superseded rule (new):* `docs/configuration.md:87` says `minIntervalSeconds` "must be kept below the requeue interval", the framing the rest of the commit replaced. It is the one place of five that was missed, and it is the argument you made yourself: at `suspendHoldRequeueIntervalSeconds=600` it claims a limit of 600 where the real one is 3000. [inline: [`docs/configuration.md:87`](https://github.com/apache/spark-kubernetes-operator/pull/861#discussion_r4073065942)]

peter-toth (approved):

dongjoon-hyun:
Thanks @peter-toth. Finding 4 fixed in 950fd4a, taking your suggestion as written.

You were right that it was a fifth site rather than one of the four I counted, and right about the direction of the error. The repo's own configuration is the counterexample, as you say: the E2E sets `suspendHoldRequeueIntervalSeconds=10` with `minIntervalSeconds=15`, which the row declared illegal even though the resulting 20 second period sits against a 3600 second TTL, so the row forbade a setting this PR itself relies on rather than merely being conservative.

Dropping the clause rather than restating the corrected bound, per your suggestion: the row already links to `#event-frequency`, and that restatement is what drifted in the first place. `grep` for the superseded framing now comes back empty, and the corrected bound is stated in six places — the `Event frequency` section, both option descriptions and the two generated `config_properties.md` rows.

`gradle build -x test` is green, `validateConfPropsDocUpToDate` included. Docs-only change, so no test run.

dongjoon-hyun:
Merged to main

